PR #2, opened 2018-06-04
Allow spaces in executable path

Brett208:
- Properly parse the executable's path to not appear as multiple command arguments.
 - Improve error message when malformed command line arguments are encountered.

### DIFF
--- a/ConsoleModuleLoader.cpp
+++ b/ConsoleModuleLoader.cpp
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <iterator>
 #include <filesystem>
+#include <stdexcept>
 
 namespace fs = std::experimental::filesystem;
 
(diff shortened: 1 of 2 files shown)


Brett208:
The windows function CommandLineToArgvW does not exist for non-Unicode strings. This required converting the returned LPWSTR container into a std::strings via a helper function. 

Tested on Windows. It allows adding spaces within the path to Outpost 2's executable without displaying an error message. It also successfully loads a command line module even with spaces in the directory path.

-Brett

ConsoleModuleLoader.cpp
```diff
@@ -90,15 +90,22 @@ void ConsoleModuleLoader::ParseCommandLine(std::vector<std::string>& arguments)
...
+						" command line argument from LPWSTR to char*. Further parsing of command line arguments aborted.");
+					break;
+				}
+				arguments.push_back(argument);
 			}
-			arguments.push_back(argument);
+		}
+		// Catch and STL produced exceptions.
+		catch (std::exception& e) {
+			PostErrorMessage("ConsoleModuleLoader.cpp", __LINE__, "Error occurred attempting to parse command line arguments. Further parshing of command line arguments aborted. Internal Error: " + std::string(e.what()));
```
DanRStevens: Typo: "parshing"

I don't think `e.what()` needs to be wrapped by a `std::string`.

This seems to duplicate error code. I'm wondering if the inner PostErrorMessage should become a throw.

ConsoleModuleLoader.cpp
```diff
@@ -90,15 +90,22 @@ void ConsoleModuleLoader::ParseCommandLine(std::vector<std::string>& arguments)
...
+			LocalFree(commandLineArgs);
```
DanRStevens: Ahh, right, I forgot about that.

Err, wait. It's already freed outside of the catch block. In the case of an exception, it will be double freed here (which is likely in the realm of undefined behaviour).

I think if you removed the `LocalFree` from inside the catch block, you should be fine. The explicit try/catch should keep your code exception safe.

An alternative might be to use [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) with the `LPWSTR *commandLineArgs` variable:
```
SomeLocalFreeCallingRaiiObject<LPWSTR*> commandLineArgs = CommandLineToArgvW(GetCommandLineW(), &argumentCount);
```

Though you'd probably have to write such an object. Such an object would need a constructor to support initializing the internal pointer, a destructor which calls LocalFree on the pointer, and a way to dereference the stored pointer, perhaps using an `operator T*()` [type conversion operator](https://en.cppreference.com/w/cpp/language/cast_operator).

Don't know if you want to take it that far though. I think simply removing the extra `LocalFree` call will make the code correct. Ahh, the joys of trying to reason about manual memory management.

Brett208: I think using RAII in this case would be a good learning experience. Perhaps this could be a small future project as it would be reusable in other work. I just fixed it for now so the code is correct though.

I just committed a change on the master to remove the extra call to LocalFree. Thanks for catching this even though it took so long to circle back to fixing it on my part.

-Brett

PR #4, opened 2018-11-30
Add message logger for use with external modules

Brett208:


Brett208:
Oh, 2 more things:

The function gmtime is considered deprecated. I tried a bit to get gmtime_s to work. Unfortunatley, I get the following error argument of type "__time64_t *" is incompatible with parameter of type "tm *" for the first argument into gmtime_s. The type `__time64_t` is the return from `std::chrono::system_clock::to_time_t(currentClock)`.

Warning	C4996	'gmtime': This function or variable may be unsafe. Consider using gmtime_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	op2ext	\op2ext\logger.cpp	33	

The output into the logger will look like this:

```2018-11-30 03:01:52 UTC [NetHelper] Net Helper Process Attached```

I think this looks fine, but we can also put some thought into if we want to create the log file in a way that easily parses into something like csv or xml format.

-Brett

File: Logger.cpp [hunk 1]
```diff
@@ -0,0 +1,36 @@
+#include "Logger.h"
+
+#include "GlobalDefines.h"
+#include <cstddef>
+#include <chrono>
+#include <ctime> //gmtime
+#include <iomanip> //put_time
+#include <sstream> // stringstream
+
+Logger::Logger() : logFile("Outpost2Log.txt", std::ios::app | std::ios::out | std::ios::binary)
```
Arklon1: I feel like this filename string should be defined elsewhere, and maybe even configurable. Separate log files per-module might even make sense if a mod might ever expect to be highly verbose.

DanRStevens: That's a good point. Should we limit scope?

The default log filename could probably become a named constant.

Brett208: I'm not sure we have need for multiple log files, but below is my first stab at what this would look like:

We could place the log file name in Outpost2.ini.
```
[Game]
...
LogFilename = "Outpost2Log.txt"
...

[NetFix]
...
LogFilename = "Netfix\NetFixLog.txt"
...
```

Perhaps the following rules:

1. If The LogFilename is not defined in the **Game** section of Outpost2.ini, Outpost2Log.txt is used.
2. If a submodule does not include a LogFilename, it maps to the master log file.
3. If a submodule includes a LogFilename, it will map to a separate log file. The default location for this log file would be the directory containing its DLL, but not an enforced rule.
4. The default log file extension should be .txt, but not an enforced rule.
 
Do we want to add a modal dialog box warning if a LogFilename is ill formed, or cannot be opened for any reason? Other option would be to fail silently or for the case of a separate submodule, log in the master log file.

We wouldn't have to label each log entry with the submodule name if the submodule has its own log, but it might be nice to leave it in to simplify the code base. It would also allow easy recognition of which log file you had open I suppose.

This would make the implementation a fair amount more complex than it currently is.

-Brett

DanRStevens: Assuming we want to allow multiple log files, the above design seems sound.

But do we need multiple log files, and if we implemented it, would that just add to the clutter and confusion? If something goes wrong, and we need to help someone debug, it might be easier if there was a single log file containing everything they could send for analysis.

In terms of verbose logging, I assume that would mainly be done during development, or when an actual problem has occurred. For release builds, standard logging could be set to a minimum level (possibly through a default configuration option). For error logging, we do want that to be complete and reasonably verbose so the problem can be diagnosed easily. I don't see much need to protect against verbose logging by splitting log files for releases.

Brett208: Hmm, I would probably agree with Dan that we don't really have need for multiple logs or even to really rename the primary log file. I can't really think of a good use case for wanting to rename the log file. 

Also, if we allow customizing the filename & location of the log file, it creates a security concern. Someone could try to name the log file ../../../system/mscorlib.dll or something which could overwrite things.

-Brett

DanRStevens: True, though if they can write to the Outpost2.ini settings, and Outpost 2 was able to write to a system folder, the computer's security is probably such that they could have just written to the system folder directly. This would mainly be an additional security concern for people accepting setting files from an untrusted person over the internet, and installing it. In that case, they probably have bigger security concerns, such as the module loader code being used to load an arbitrary DLL which could do whatever.

Having a standard name does simplify troubleshooting instructions though.

DanRStevens:
Alright, I was looking into the `gmtime` thing.

----

First up, are we even using the right set of functions to accomplish what we want to do? I noticed some of those methods are rather old C methods, using C header includes, with old style pointer usage and error checking. I went looking for a more modern C++ way of doing things.

I found the following summary of the [`std::chrono`](https://en.cppreference.com/w/cpp/chrono) library interesting. Unfortunately, most features listed are slated for C++20. Even using GCC 8, and enabling experimental C++20 features with "-std=c++2a", they're still not available. In particular, [`utc_clock`](https://en.cppreference.com/w/cpp/chrono/utc_clock) is not yet available. A shame, as it would have supported serializing UTC time directly to a stream with `operator <<`. One can always hope for the future.

After much searching, it doesn't look like there is a current C++ way of working with time without using old style C functions.

Curiously, I came across the following note on the [`std::chrono:system_clock`](https://en.cppreference.com/w/cpp/chrono/system_clock) page:

> It is the only C++ clock that has the ability to map its time points to C-style time, and, therefore, to be displayed (until C++20).

----

I looked into [`gmtime` and `gmtime_s`](https://en.cppreference.com/w/c/chrono/gmtime). The `gmtime_s` variant accepts an additional output buffer parameter `tm* result`. The original likely used an internal static buffer, and then returned a pointer to the static buffer. That would have threading and re-entrancy issues, hence the alternative "_s" method. The [`tm` struct](https://en.cppreference.com/w/c/chrono/tm) holds individual year, month, day, hour, minute, second, and other fields, rather than storing the time as an offset in seconds from the Epoch. The `gmtime` functions do the conversion from seconds to unpacked fields.

One unfortunate point I barely noticed in documentation, is that `gmtime_s` function is from the C11 standard, not the C++ standard.

Likely related to the standard note above, I've been encountering some error when trying to experiment with `gmtime_s`. In particular, the function declaration shown by the compiler in the error message, does not match the signature from the documentation.
Error message: `errno_t gmtime_s(tm*, const time_t*)`
Documentation: `struct tm *gmtime_s(const time_t *restrict time, struct tm *restrict result);`
You'll notice the parameter order is swapped, and the return types are different.

----

That's all for now.

Brett208:
After merging the changes in this branch into the child auto-module name lookup branch, I tested without any discrepancies. The time format for a CPU set to locale English US produces the same results. There are pros and cons to formatting both ways. I think standardizing across regions for consistent results is probably a good choice overall. Especially considering the function is talking about getting the system time.

I'm ambivalent to where UTC is appended. I had meant for GetSystemDateAndTime to be generic, so perhaps there are use cases where you do not want UTC appended? Since it is currently a member of Logger.cpp (well I guess it isn't a member function anymore but lives in the same file) it is certainly appropriate to append within GetSystemDateAndTime. 

I would support moving GetSystemDateAndTime into a separate file, but we do not have a use case for it outside of the logger so probably not worth it at this time. 

Your thoughts on the standard library timing functions being lacking at best and forcing an odd mix of C++ and C includes pretty much match what I encountered. Frustrating there isn't a simple way to pull the system time and format reasonably within the C++ standard library. Makes me appreciate DotNet Framework's extensiveness and user friendliness at times like this.

DanRStevens:
Indeed, it is a frustrating mix of antiquated C methods. I'm glad C++ is finally getting some much needed improvements in the recent standards, though the changes are sometimes painfully slow, and often long overdue.

As for where UTC is appended, I figured since the function is hardcoded to return UTC time strings, it was probably most appropriate there. I suppose the time format string could be parameterized though, with a default value. I would also support moving it to it's own file. I was just too lazy to do it at the time. And yeah, it does seem standalone, and potentially usable in other areas of the code.

----

I went ahead and updated the code to use the Microsoft documented version of `gmtime_s`. Mingw accepted the code without error, and AppVeyor hasn't complained about it. The error handling is slightly ugly though. I went with an approach similar to that used in the module name lookup code.

DanRStevens:
Looks like the warning has been resolved. Assuming this works, I recommend merging it.

Brett208:
Just tested by compiling in both debug and release mode. In both modes, the module is able to properly post a log and op2ext is able to find the module's name automatically.

I narrowed down a separate bug associated with custom volume loading that I'll post about in a separate branch. 

Thanks for the work on this branch. I think it came together nicely and I wouldn't have been able to figure out gmtime_s or the module lookup on my own.

PR #7, opened 2018-12-11
Add auto module name lookup

DanRStevens:
Ok, I gave the initial implementation a try. This will need testing. I need to think about how this can be tested, but meanwhile, I thought I'd share it.


DanRStevens:
Ahh, thanks for getting the bits I missed. In the future I'll have to remember to run a quick compile with mingw for *.cpp. For some reason, I didn't think of that after realizing there was no Makefile. Which is odd, because I had enough sense to run the new code through mingw to check for errors.

The module detect code should work for any module, including op2ext. Should be no need for special code.

Brett208:
Okay,

You did the heavy lifting in the win32 programming so I was happy to clean up the easy bits.

Just tested this branch. The following log is output when using the console module loader on the test module:

```2018-12-13 03:30:04 UTC [<Module not found>] Test adding a message to log file```

I output all loaded modules, and the following are loaded:

Outpost2.exe
ntdll.dll
KERNEL32.DLL
KERNELBASE.dll
USER32.dll
win32u.dll
GDI32.dll
gdi32full.dll
msvcp_win.dll
ucrtbase.dll
ADVAPI32.dll
msvcrt.dll
sechost.dll
RPCRT4.dll
SspiCli.dll
CRYPTBASE.dll
bcryptPrimitives.dll
ole32.dll
combase.dll
WINMM.dll
COMCTL32.dll
WSOCK32.dll
WS2_32.dll
OP2EXT.DLL
SHELL32.dll
cfgmgr32.dll
shcore.dll
windows.storage.dll
shlwapi.dll
kernel.appcore.dll
profapi.dll
powrprof.dll
FLTLIB.DLL
WINMMBASE.dll
MSVCP140D.dll
VCRUNTIME140D.dll
ucrtbased.dll
IMM32.DLL
op2mod.dll

(That is a lot of modules...)

The correct module should have been op2mod.dll. That is the name of the test module.

The code is returning right after the do loop is completed. So it cycles through all of the modules and does not locate the proper one.

```C++
if (Module32First(hModuleSnap, &moduleEntry)) {
	do {
		OutputDebug(moduleEntry.szModule);
		OutputDebug("\n");
		if (containsAddress(moduleEntry, address)) {
			return std::string(moduleEntry.szModule);
		}
	} while(Module32Next(hModuleSnap, &moduleEntry));
	return std::string("<Module not found>"); // RETURNS FROM THIS LINE
}
```

Tested in both release and debug mode without change.

-Brett

Brett208:
Also, get the same message when attempting to write from op2ext.dll:

`2018-12-13 03:39:50 UTC [<Module not found>] Test from op2ext`

DanRStevens:
You should dump the return address and see if it is sensible. Quite likely the offset to the return address is wrong. You may need to update the offset in the line:
https://github.com/OutpostUniverse/op2ext/blob/9d9b16475089fc392694a2df1bbda19ae11c513b/Logger.cpp#L25

If it helps, you can also dump the address of the `char* message` parameter. The return address should be somewhere on the stack between the local variable `void* returnAddress` and the parameter `char* message`.

Brett208:
I may not have time to work on this for several days. Sorry. I've never done this sort of debugging before so I will attempt when have some more time but may want some extra help.

-Brett

DanRStevens:
If you can compile the code to an assembly listing, the assembly listing can be analysed to pull out the right constant.

Failing that, you could potentially guess and check:
```cpp
void* returnAddress = static_cast<void*>(&returnAddress + 1);
void* returnAddress = static_cast<void*>(&returnAddress + 2);
void* returnAddress = static_cast<void*>(&returnAddress + 3);
...
```

The bounds for the address of the function's return address is between `&returnAddress` (the local variable) and `&message` (the parameter). This range is not normally very large, though depends on the number of parameters, local variables, and temporaries (effectively unnamed local variables). The method in question uses an `std::string` which could be of a fairly large size.

ASCII Art:
```
(high address)
-----------------  ----  [Parameters]
| message       |  char* message
-----------------  ----  [Function return address, and other bookkeeping]
| codePtr       |  (Function return address)
-----------------
| framePtr      |  (Base/frame pointer, assuming no "omit-frame-pointer" optimization)
-----------------  ----  [Locals and temporaries]
| returnAddress |  void* returnAddress
-----------------
| moduleName    |  std::string moduleName
-----------------
| (temporaries) |  (maybe)
-----------------  ----
(low address)
```

The exact order of the variables is of course subject to the whims of the compiler and it's optimization settings. Though placement is typically in declaration order, unless there was a good reason to modify it.

DanRStevens:
Ahh, excellent, very helpful.

----

In case you're curious...

The beginning section is where it sets up the stack frame. There is more in the stack frame than I realized. In particular, there is quite a lot of space reserved for locals/temporaries/hidden security related fields. After this, it initializes the variable of interest:
```asm
void Logger::Log(const std::string& message)
{
10011DE0  push        ebp  
10011DE1  mov         ebp,esp  
10011DE3  push        0FFFFFFFFh  
10011DE5  push        1001B556h  
10011DEA  mov         eax,dword ptr fs:[00000000h]  
10011DF0  push        eax  
10011DF1  sub         esp,54h  
10011DF4  mov         eax,dword ptr [__security_cookie (10024038h)]  
10011DF9  xor         eax,ebp  
10011DFB  mov         dword ptr [ebp-10h],eax  
10011DFE  push        esi  
10011DFF  push        eax  
10011E00  lea         eax,[ebp-0Ch]  
10011E03  mov         dword ptr fs:[00000000h],eax  
10011E09  mov         dword ptr [this],ecx  

		void* returnAddress = static_cast<void*>(&returnAddress + 1);
10011E0C  lea         eax,[ebp-50h]  
10011E0F  mov         dword ptr [returnAddress],eax
```

The methods starts off by setting up a new stack frame, which contains the local data for a particular function invocation. The base pointer marks the fixed location at the boundary between the caller's stack frame, and the callee's stack frame. The stack pointer marks the other end of the callee's stack frame, which can move about as scope blocks are entered or exited, or as parameters are pushed onto the stack for future function calls. The standard start to a method is saving the old base pointer (`push ebp`), and then moving the current stack pointer into the base pointer (`mov ebp, esp`).

Next there's a few `push` instructions, which add data to the stack frame. The references to `fs:[00000000h]` is accessing the current Structured Exception Handler (SEH) frame field of the Thread Information Block (TIB). The `fs` register is used for thread local storage in Win32. The `push 1001B556h` is likely storing the address of the exception handler routine. You'll notice the address is similar to the code addresses in the left hand column.

You may notice the `sub esp, 54h`, which reserves space for locals (and temporaries). The `esp` register is the 32-bit Stack Pointer, and it grows down, hence subtract.

The function prologue must also save the contents of certain non-trashable registers it intends to use, hence `push esi`. The non-trashable registers for Win32 programming are `ebx`, `ebp`, `esi`, and `edi`. The other 4 registers `eax`, `ecx`, `edx`, and `esp` are considered trashable, where a function may use them freely without saving/restoring their previous contents.

You may notice there is also a `push eax` instruction. This is not related to preserving register contents. Above you can see `eax` is loaded with the `__security_cookie` and `xor`ed with the base pointer. This is part of a security mechanism to protect against stack based buffer overflow exploits. At the end of the method, there is a call to `__security_check_cookie`, which should validate these fields, before allowing the method to return. Any discrepancy, and it's assumed the program has been compromised and is terminated immediately.

Due to all the extra bookkeeping on the stack, the distance from the `returnAddress` local variable, and the actual return address value is quite large. In the instruction `lea eax, [ebp - 50h]`, it is calculating `static_cast<void*>(&returnAddress + 1)`. Effectively this is `eax = ebp - 0x50 = ebp - (localVariableOffset(returnAddress) + 0x4)`. From this, we might infer the local variable `returnAddress` is at offset `-0x54` (the stack grows down) from the base pointer within the local stack frame.

Side note: Normally the `[expression]` represents a read from a memory address denoted by expression, however, the Load Effective Address (`lea`) instruction is actually an arithmetic instruction which can do combined shift and add. It's use is to support array indexing, such as `lea destRegister, [baseAddress + scale * index]`, where scale can be 1, 2, 4, or 8. As this general form of computation is often useful outside of memory address calculations, it is often used in other contexts.

----

Ok, so given the above, perhaps try either of:
```cpp
void* returnAddress = static_cast<void*>(&returnAddress + 22);
void* returnAddress = static_cast<void*>(&message - 1);
```

The later might be simpler and easier to understand from a maintenance perspective. There will be less on the stack between the parameters and the return address then between the locals and the return address.

Brett208:
I added a commit that marks the passed `void*` as `const void* const`. I thought this would add flexibility to what was passed into WindowsModule.h and provide a little safety on us actually modifying the pointer.

I tried adding &returnAddress + 22 but still received a <Module not found> message. Tried in both debug and release compilations without success. Maybe your math is a little wrong on getting to 22? Does attaching the debugger change the formula?

I tried &message - 1. The compiler spit out the following warning: static_cast throws an invalid type conversion on &message - 1. By switching to a C-style cast () instead of static_cast, it would compile although not sure why the static_cast would not work.

The automated build will keep failing until we merge in the appveyor branch.

Thanks for the explanation on assembly. I appreciated reading through it. My knowledge of assembly is so limited that it doesn't necessarily help me solve the problem at hand but it is interesting. Before you posted, I had poked around the internet reading about what EAX means. The answer seemed complicated because its use had changed between x16 and x32 (I think?). 

-Brett

DanRStevens:
Ahh, because the `message` parameter was marked as `const`, and `static_cast` can not remove `const`.

----

Good call on marking the address as `const`. Though I think the `const` is most appropriate on the pointed to value, rather than the parameter itself. I would have used `const void* address` there. Using `const` on the pointer itself (`void* const`) seems a bit silly in that same sort of way that declaring a parameter as `const int byValueParamMarkedAsConst` seems odd.

Have you tried omitting the second `const` in the header file and method declarations, but retaining them in the .cpp implementation? I tested a bit on Linux with Mingw, and that seemed to work.

----

It's entirely possible my math was off on this one. Hopefully my guesses are at least in the right ballpark now.

----

For AppVeyor, you could create some manual settings, such as specifying the solution file to build. It should be all configurable from the web.

----

Can talk more about assembly another time. :grin:

Brett208:
While I generally understand the definition of a making a pointer const (cannot change the address it points to), I'm not really familiar with using it in practice. So I just removed them.

I just tried the following code block, which was unable to find the module's name. So I feel like there is something wrong still...

``` C++
for (int i = 0; i < 200; ++i)
{
	void* returnAddress = static_cast<void*>(&returnAddress + i);
	std::string moduleName = FindModuleName(returnAddress);
	logFile << GetSystemDateTime() << " UTC [" << moduleName << "] " << message << std::endl;
}
```

I merged in the appveyor settings so this should now 'pass'. I accidentally deleted this branch using TortoiseGit on my computer. That caused it to close, etc. I reopened everything. Sorry for the noise.

-Brett

DanRStevens:
Hmm, interesting. There might be another bug in the code then.

Maybe add some debug output to display the return address. Would also be good to see the base address and size (or end address) or each module.

This might be a good candidate for a pair programming session, as it's hard for me to debug from my Linux machine.

----

On another note, I noticed I was a little inconsistent about parameter order with `returnAddress` in WindowsModule.h. Maybe the parameter order should be adjusted for consistency.

DanRStevens:
Oh, duh, I didn't dereference the pointer:
```cpp
std::string moduleName = FindModuleName(*returnAddress);
```

Normally I'd have used a reference here, though I'm not sure you can use a void reference.

DanRStevens:
Mingw can't handle the `#include <filesystem>`, hence I haven't been doing a full compile, nor been able to run any of the code.

Maybe some more details on the test code could help. The return address can be affected by function inlining. That could cause it to report the wrong module. Though I don't expect any inlining of the `Log` method to occur here.

Curiously though, Mingw had compiled `_ReturnAddress()` into a `call` instruction, which I thought was odd, and would likely give bad results if a generic method was used. I would very much expect this intrinsic to be inlined. Wasn't sure if that was simply due to bad support for the intrinsic in Mingw though.

Getting a module name is promising though. Maybe we could go back and try:
```cpp
const void* returnAddress = *static_cast<const void**>(&returnAddress + 22);
std::string moduleName = FindModuleName(returnAddress);
```

Mingw seems to produce the right form of assembly code for that, though the offset doesn't match the stack frame that Mingw generates. That's one reason why being able to use an intrinsic would be nice. The intrinsic should automatically adjust for changes in the stack frame layout.

----

As a debugging measure, it might be good to dump the address of the `Log` method, the address of the calling method, and the value of `_ReturnAddress()`/`returnAddress`. The three should all be different. In correct code, the `_ReturnAddress()` should be a reasonably small offset beyond the address of the calling method. I suspect what is happening is it's giving a value that's a small offset beyond the `Log` method instead.

Brett208:
Reading through your post, I had an epiphany. We are wrapping Logger::Log in the exported C ABI function Log (within op2ext.cpp). So, Logger::Log is returning a result to the op2ext.cpp log function whenever it is called from an external module. 

If I call FindModule in the log method within op2ext.cpp, a proper result of op2mod.dll is returned.

-Brett

DanRStevens:
Aha! So it was something simple like that. I forgot about the C wrapper.

Ok, so we need to grab the return address within the public interface method that is exported from the DLL. Any deeper in the call stack, and we'll get return addresses within op2ext itself.

I suppose that means the `Logger` class interface needs to be updated to accept the return address as an additional parameter. Optionally, we could provide an overload if we want to also support direct calls from C++, though I don't think we need to do this.

We might want to rename the parameter from `returnAddress` to something else. Effectively, the return address was serving as a unique identifier (in the absence of [tail call elimination](https://en.wikipedia.org/wiki/Tail_call)) for the log message source. It was unique to more-or-less the line of source code generating the message, and we were simplifying by using the identifier to lookup the module of the log message source. Maybe we could name the parameter `void* messageSource` or similar.

Brett208:
Glad we figured this out. I had been thinking about recommending abandoning this since we couldn't seem to get it working.

Since op2ext.h is our public include, I was trying not to call it from within the op2ext codebase. Maybe this is okay to do, but it felt a bit like a circular reference. From within op2ext, I think you would include  op2ext-Internal.h which contains the logger class for logging purposes.

I think exposing a new log function within op2ext makes sense here. It avoids the module lookup when not needed and prevents having to cast the message string into a char* to pass it into op2ext.h only to cast it back to a string for use by the logger.

I really like using the _ReturnAddress function. I was worried about how we could create a compilation error if compiler settings altered the exact integer offset we were applying to the returnAddress pointer. Using _ReturnAddress solves this.

messageSource makes sense to me. Should probably still  document what messageSource is with a comment as this isn't a trivial application.

I won't be working on it at all until tomorrow so feel free to make changes if you want.

-Brett

DanRStevens:
Ok, here I the changes I think we need. No need to modify op2ext.h, only op2ext.cpp.

Guess I'll have to think a bit about documentation, though the return address thing will still be internal to the library, so module writers won't need to worry about it.

DanRStevens:
Good call on the const.

Do you have an opinion on where the `FindModuleName` call should be? I was looking at the updates, and wondering if that should be moved to the exported C API method. The logger could then be simplified by removing the method with the `messageSource` parameter, and making the private method with the `moduleName` parameter public. That might also make for a more understandable API.

Going with the above, we could also further reduce the `Log` class API surface by adding a default value to the `moduleName` parameter, and removing the other methods.

The other alternative I thought of, was to have the single parameter `Log` method using `_ReturnAddress()` internally, and delegate to the method with the `messageSource` parameter, which would eliminate the need for the private method. Fewer methods, but this would seem to be less efficient than your solution. Another ugly point is this proliferates the use of the `_ReturnAddress()` intrinsic, which I think we should keep to a minimum. I was also considering security of accepting a `moduleName` parameter, though you could lie about the `messageSource` parameter just as easily as you could lie about the `moduleName` parameter, so no real security difference.

Ehh, now I'm just thinking out loud. I'll try the first idea.

----

I think the `gmtime` warning can be addressed in a separate branch.

Reminder: This branch is based on the original logging branch, not on master. Hence, we merge this branch into the logging branch, and then later merge the logging branch into master.

Brett208:
In order to test the features added to this branch and the parent branch, I merged the parent branch into this one. It was fairly difficult due to all the changes made on top of each other. Hopefully I didn't miss anything. 

I'm happy with the changes made to simplify the logger public interface. I guess it didn't bother me either way. I'll comment on the other changes in the other branch.

-Brett

DanRStevens:
Seems you beat me to it. I noticed I'd created a few merge conflicts with the base branch. Amusingly, when comparing things back to master, about the only thing that actually changed in the `Logger` class, was the parameter order to the `Log` method. All the other additions and changes were basically undone or moved elsewhere.

Alright, time to move forward. I'm going to merge this to the base branch now.

PR #9, opened 2018-12-14
Use malloc instead of new to allocate in external C API function GetCurrentModDir

### LINKED ISSUE #8: Error in op2ext TestModule in function TestGetConsoleModuleDirectory
When loading the test console module, MSVC seems to be reporting an error occurs when the command free() from below is run. Not sure what I'm doing wrong to cause this.

``` C++
// GetCurrentModDir only returns the directory of a module loaded through the console switch /loadmod
void TestGetConsoleModuleDirectory()
{
	char* moduleDirectory = GetCurrentModDir();
	std::string modDirectoryReport("GetCurrentModDir reports: " + std::string(moduleDirectory) + "\n");
	OutputDebugString(modDirectoryReport.c_str());

	free(moduleDirectory);
}
```

Brett208:
This fixes a legitimate bug in op2ext's external API. Perhaps we should role it into Outpost 2 before the next release of the game?

Also, not sure if we want to use static_cast there or c style cast. I think it is okay to call the static_cast here since the cast doesn't affect what is passed outside of the function and across DLL boundaries. 

Closes #8

### DIFF
--- a/op2ext.cpp
+++ b/op2ext.cpp
@@ -47,7 +47,7 @@ OP2EXT_API char* GetCurrentModDir()
 		return nullptr;
 	}
 
-	char* cStr = new char[modDirectory.length() + 1];
+	char* cStr = static_cast<char*>(malloc(modDirectory.length() + 1));
 	strcpy_s(cStr, modDirectory.length() + 1, modDirectory.c_str());
 
 	return cStr;


DanRStevens:
I was thinking about this, and there may be another problem. The op2ext module is compiled separately from whatever client code that uses the returned string. As such we've crossed a module boundary between the point of `malloc` and the point of `free`. As the code in each module won't necessarily link to the same methods and global data structures, this can also result in undefined behaviour. In short, don't mix heap operations between modules.

The proper way to do this, is likely to provide a Free method, which the calling code could pass the string to after it is done with it. The above is the reason why most libraries with dynamically allocated resources also provide a corresponding `Free*` method.

This is also why a lot a APIs allow you to pass in a buffer to write into, which greatly simplifies memory management. However, then the called code is unable to resize the buffer, and so if the string is too big, it has to fail and somehow signal the error.

Brett208:
op2ext's public API is a fair mess overall considering how small it is.

When we went with GetGameDir_s, we used the buffer + size concept. It might be nice to stick with that here for consistency, but that would require deprecating the current function.

Maybe creating a free function would keep from deprecating, which would be nice.

Dan, do you have a preference? I'm not knowledgeable enough to care either way as long as it works.

Will probably abandon this branch and create a new branch when we chose a way forward.

Thanks,
Brett

Brett208:
Okay, well I guess we could mirror the GetGameDir_s call and create a GetCurrentModDir_s function. Below is the GetGameDir_s syntax. This would allow calling larger than max_path although a user could start with max_path for the buffer size.

We could then mark the old function as deprecated.

// Retrieves the current absolute directory of the Outpost 2 executable with a trailing slash. 
// If bufferSize is smaller than required to copy entire path, buffer is provided as much of path as possible.
// Returns 0 on success. Returns the required minimum size of the buffer on failure.
OP2EXT_API size_t GetGameDir_s(char* buffer, size_t bufferSize);

DanRStevens:
Alright, let's go with that.

Brett208:
I didn't voice this earlier, but my concern is if previous modules have employed the function knowing that it uses new instead of malloc, will we be causing harm in those DLL's memory management by switching it?

If someone wants to investigate, console modules that are probably worth reviewing would be cm1, cm2, and the tech tree mod that I forget the name of.

If you think it is still the right choice, then let's merge in.

Restoring branch until we make a choice.

-Brett

DanRStevens:
It's hard to say how it would affect them without looking at both the source, and what the compiler did with it.

Though if they ignored the problem, and just leaked memory, then it really doesn't matter how we allocate it. That might very well be the case.

I'm uncertain where to find the source for those projects.

Brett208:
Arklon might have the source to cm1 and cm2. sirbomber would have the source for the other one. Would be nice to have both in the repository.

I could load the modules and set a breakpoint in the function GetCurrentModDir within op2ext. Then by loading Outpost 2 and using the mod for a bit, it should tell us if they are loading using the function or not. I haven't tried this before, but if the breakpoint triggers, then VS may let me see the assembly associated with the call across the module sort of like what I provided when we were troubleshooting.

Below is the complete list of mods I know about. It would be really nice if someone could take the time to test and document them all. Some sort of a la carte menu for people to download and mod the game with. Unfortunately if I recall correctly, many are broken.

 - Color Mod 1 (CM1) - sirbomber/Arklon (console loaded)
 - Color Mod 2 (CM2) - sirbomber/Arklon (console loaded)
 - Multitek2 - sirbomber (Console loaded, depends on ART_PATH only)
 - NetFix - Hooman (.ini loaded)
 - NetHelper - Arklon (.ini loaded)
 - Renegades - EddyB and team
     - It uses the Console module loader to work. However, the installer mentions that it will install a patch to fix a bug in the console module loader (old name modmgr). No idea what the bug is though.
 - OP2DisasterlessMod - Arklon (Ini/Console)
 - OP2GPPatch - ZigZagJoe (Ini/Console)
 - OP2ScrollFix - ZigZagJoe (Ini/Console)
 - SSUtil2 (Screenshot Utility) - TH300 (Ini/Console)
 - Video Recorder (Arklon has source) Contains major bugs

After thinking about this some more, I think we should just leave the deprecated function alone. If someone properly cleaned up the new call, I don't want to break their work. If a new module is compiled, it needs to use the new function. Testing all these modules seems to be a lot of work for a deprecated function.

-Brett

DanRStevens:
Ehh, that would be a major pain to track down source and test things. Particularly since I haven't heard from some of those people in years. Last I spoke to Eddy, he'd lost access to the Renegades source. And the method, with or without the patch, is broken anyway.

As much as I don't like to just leave things, I think you might be right. This probably isn't worth the effort to pursue further. Let's just close this.

PR #11, opened 2018-12-19
Add Appveyor automated build process

Brett208:
- Update post build script to work in automated build process.

The post build script had to be updated. xcopy requires input on whether the source is a directory or a filename when the directory does not exist, hence the `echo D|` line. I hadn't noticed until now since the Outpost2 directory already existed in my development process. This post build event doesn't add anything to the appveyor build process, but I didn't see an easy way to turn it off only in the appveyor build so figured it was fine to let it run anyways.

-Brett

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,11 @@
+image: Visual Studio 2017
+configuration:
+  - Debug
+  - Release
+platform:
+  - x86
+install:
+  - cd %APPVEYOR_BUILD_FOLDER%
+  - git submodule update --init --recursive
+build:
+  project: op2ext.sln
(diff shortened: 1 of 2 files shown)


Brett208:
Nice. Why didn't I think of that. Ammend commit with force push, so delete branch to update if you pulled it.

-Brett

DanRStevens:
Ahh neat, GitHub marks the forced-push in the pull request conversation history. I hadn't seen that before, and I've done a number of forced pushes in the past. I wonder if it's new, or if I just hadn't noticed before.

My local Git copy updates just fine. There's a small note about a forced update there too, though it all works without any intervention on my part. (I only had the remote tracking copy of the branch, no checked out local copy of that branch).

Looks like the trailing directory slash works.

Brett208:
I don't remember it posting that about your forced commits, but glad to see it being highlighted. Okay, merging now.

PR #17, opened 2018-12-23
Mark input strings as const in exported methods

DanRStevens:
Noticed a small oversight when I tried to compile the test modules with Mingw. It gives a warning about how ISO C++ forbids automatic conversions from `const char*` to `char*` when the call sites try to pass a string literal (which is a `const char*`).

### DIFF
--- a/op2ext.cpp
+++ b/op2ext.cpp
@@ -65,7 +65,7 @@ OP2EXT_API char* GetCurrentModDir()
 	return cStr;
 }
 
-OP2EXT_API void AddVolToList(char* volFilename)
+OP2EXT_API void AddVolToList(const char* volFilename)
 {
 	if (modulesRunning) {
 		PostErrorMessage("op2ext.cpp", __LINE__, "VOLs may not be added to the list after game startup.");
@@ -88,7 +88,7 @@ OP2EXT_API void SetSerialNumber(char major, char minor, char patch)
 	}
 }
 
-OP2EXT_API void Log(char* message)
+OP2EXT_API void Log(const char* message)
 {
 	// The return address will (normally) point into the code section
 	// of the caller. The address can then be used to lookup which
--- a/op2ext.h
+++ b/op2ext.h
@@ -48,7 +48,7 @@ OP2EXT_API char* GetCurrentModDir();
 // initializes. Typically should be called in module's mod_init (InitMod) function.
 // @param volFilename The vol filename ending in a null terminated string. 
 //                    Must be a relative path from the directory containing the Outpost 2 executable.
-OP2EXT_API void AddVolToList(char* volFilename);
+OP2EXT_API void AddVolToList(const char* volFilename);
 
 
 // Overwrites the default Outpost 2 version string.
@@ -57,7 +57,7 @@ OP2EXT_API void AddVolToList(char* volFilename);
 OP2EXT_API void SetSerialNumber(char major, char minor, char patch);
 
 // Log a message in Outpost2Log.txt.
-OP2EXT_API void Log(char* message);
+OP2EXT_API void Log(const char* message);
 
 #ifdef __cplusplus
 } // extern "C"


DanRStevens:
Side note: This change is binary compatible.

I did however just manage to create a link error by only updating the declarations, and not the definitions. This wasn't caught during either compile or link of the library. It was only caught when trying to link the test module with the library. I think I know why though.

Changing the function signature can change the decorated name, and so by having two distinct function signatures I may have inadvertently created a parallel set of functions. However, these are `extern "C"` exports, and so the name decoration will be stripped. This relates to how C++ allows for function overloads (which require names decorated with type information to distinguish them), while C does not support function overloads, and uses undecorated symbol names. In C, a difference needs to be baked into the name itself to differentiate it.

The mistake seems to be related to the use of `extern "C"` in the header file, but not the implementation file. If the function signature matches that of the header file, the `extern "C"` declaration applies, without having to specify it again on the function definition. However, if the signature does not match (and changing a parameter to `const` is a change of signature), then it creates a new C++ overload, which does not have `extern "C"` applied to it, and so the symbol name is decorated. The library would then have an undecorated C export (but no definition), and a decorated C++ export (but no header declaration), with different symbol names. As the symbol names are different, there is no redefinition error.

I just tried adding `extern "C"` to the implementation file, and removing the `const`, and the error was caught early during compilation of the library. No need to wait for linking, nor compiling or linking of the test module.

----

Right now I'm thinking perhaps we should bake the `extern "C"` into the `OP2EXT_API` export macro. However, I removed `extern "C"` from that macro, without explanation, way back in: 0d6eed61

I assume there must have been some kind of warning related to this. Perhaps we should revisit it.

Brett208:
I tested this by commenting out extern "C" from the ifdef and pushing it into the macro:

``` C++
#ifdef __cplusplus
//extern "C" {
#endif

#ifdef OP2EXT_INTERNAL_BUILD  // Defined in op2ext project settings
#define OP2EXT_API extern "C" __declspec(dllexport)
#else
#define OP2EXT_API extern "C" __declspec(dllexport)
#endif

...

#ifdef __cplusplus
//} // extern "C"
#endif
```

I get the following errors:

Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	38	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	26	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	32	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	44	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	51	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	57	
Error	C2059	syntax error: 'string'	TestModule	c:\...\op2ext\op2ext.h	60	

The errors are generated by the test module due to the following lines:

``` C++
//Test external interface is compatible with C compiler.
//#include "CompatibilityTest.c"
```

We moved 'extern "C"' out of the macro to allow someone to call the functions from a file compiled in C instead of C++. It was probably just an omission on our part that the following lines were not also inserted in the file op2ext.cpp.

``` C++
#ifdef __cplusplus
extern "C" {
#endif

#ifdef __cplusplus
} // extern "C"
```

#endif

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2059?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(C2059)%26rd%3Dtrue&view=vs-2017

https://stackoverflow.com/questions/1041866/what-is-the-effect-of-extern-c-in-c

-----

I'm not exactly tracking why you were targeting just adding the const to the declaration and not the definition file in the first place. Although the resulting discuss was interesting. 

I haven't unpacked if it is safe to add const to the function signature without changing binary compatibility. If you have already researched this, then I'm happy taking your word on the issue. Is this to afford more flexibility in what someone else can pass into the public API or to document that we will not change the provided data (or both)?

Thanks,
-Brett

DanRStevens:
Ahh, right! The `extern "C"` isn't valid in C, only in C++. It was stripped out of the macro as a way to keep the code compatible with a plain old C compiler.

I suppose it could have been conditionally included in the macro, rather than conditionally wrapping the contents of the header file. Maybe I'll play around with that in a future branch.

----

As for why I had only updated the header file, that was a mistake on my part. I'd simply forgot to update the .cpp files too. It wasn't until the CI builds failed that I noticed my mistake. Though the error had me concerned for a while that perhaps something deeper was wrong. As in, why didn't my test compile catch the error, and why was the error occurring so late in the process. It made me doubt for a moment that the change was binary compatible.

And yes, the addition of `const` is both to allow the API to be more flexible in what it accepts, and to make it more self documenting that the method doesn't modify the parameter.

I think I'm satisfied with what's going on here now, and am ready to merge this.

PR #18, opened 2018-12-23
Initial update of documentation to version 2.1.0

Brett208:
Any future changes need to be rolled in before release.


DanRStevens:
> [DEPRECATED] char* GetCurrentModDir(); - Returns the directory of the console loaded module only (if one exists)

Hmm, this got me thinking. I'd forgotten about the console loaded module part. The ini module and the console module interfaces were designed at different times, and were slightly incompatible. Perhaps there is a way we could just dump the function going forward.

The ini modules got access to their configuration section name as a parameter to their `InitMod` function. They could use that to read their own configuration settings, which includes being able to read which directory the module comes from. We could potentially do something similar for console modules with a new Init method.

The direction of information flow for ini mods avoids memory allocation and cleanup issues. The info is passed in from op2ext to the module during module initialization. There is no need for the module to request it of op2ext. If it needs it, the module just stores a local copy. All memory allocation or buffering is internal to the module. Further, it knows ahead of time how much memory is needed, since it's presented with a read only copy of the full string it may want to store.

I may have been thinking about these changes from too low of an implementation perspective. I'd lost sight of the high level picture of how it all fits together.

Though with that said, I'm not exactly eager to make high level changes at this time. Regardless of changes, we're still stuck supporting old deprecated methods.

----

Looks like a reasonable summary of changes.

Brett208:
Thanks for reading through the changes. I'm going to merge now. I'll try to add anything to the readme changelog that is appropriate as we continue towards the next release. (Whenever that is).

PR #21, opened 2018-12-25
Add public max volume count property

Brett208:
Just realized that calling the test function `TestInvalidVolFileName()` adds a volume to VolList that does not exist in the directory. This is what is throwing the count off in `TestTooManyVolFilesLoaded()`. `TestTooManyVolFilesLoaded()` counts the volumes listed in the directory but doesn't understand that one is added that doesn't exist to test failure. We need to account for this volume addition to get the numbers right.

**Don't think the idea below would work or is really necessary.**
~~@DanRStevens, what do you think of adding a property~~ 

```C++
inline unsigned VolList::GetVolFileCount() const { return volFileCount; }
```

~~We would have to give the test code direct access to the VolList instance used by op2ext by including the header op2ext-internal.h in the test code.~~

~~This would allow accounting for any pretend volumes loaded into VolFile before checking the max buffer size. This is getting a bit complicated.~~

I'm done for a while, so please hack away at this branch. 

Thanks,
Brett

### DIFF
--- a/TestModule/TestFunctions.cpp
+++ b/TestModule/TestFunctions.cpp
@@ -1,6 +1,6 @@
 #include "TestFunctions.h"
 #include "CountFilesByTypeInDirectory.h"
-
+#include "VolList.h"
 #include "op2ext.h"
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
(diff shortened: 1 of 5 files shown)


Brett208:
I think this code is ready to merge now.

I believe the idea I expressed at end of initial message in this pull request was incorrect.

-Brett

TestModule/TestFunctions.cpp
```diff
@@ -56,20 +56,18 @@ void TestGetConsoleModDir()
...
-{
-	auto volsInGameDirectory = CountFilesByTypeInDirectory(GetGameDirStdString(), ".vol");
+	// Reserve enough space to load existing volumes in Outpost 2 directory.
+	// Outpost 2 will not load properly if certain files contained in volumes are not present.
+	auto loadedVolumes = CountFilesByTypeInDirectory(GetGameDirStdString(), ".vol");
+	loadedVolumes++; // Reserve space for VolumeLoadFail.vol
 
 	std::string volPath("./TestModule/TestVolume.vol");
-	for (auto i = volsInGameDirectory; i < 31; i++)
+	for (auto i = loadedVolumes; i < VolList::MaxVolumeCount + 1; ++i)
```
DanRStevens: I had to stop and think about the `MaxVolumeCount + 1` part. What if we pulled that expression out into a named constant:
```cpp
const auto exceededVolCountLimit = VolList::MaxVolumeCount + 1;
```

Mind you, I'm feeling a bit mentally blunted at present, and as soon as I read the function name it was clear.

VolList.h
```diff
@@ -22,14 +22,16 @@ class VolList
...
 private:
 	std::vector<std::vector<char>> volPaths;
 	unsigned int volFileCount;
 	VolSearchEntry* volSearchEntryList;
 	
 	// Static buffer, to avoid dynamic memory allocation before heap is initialized
-	static const unsigned int VolSearchBufferSize = 32;
-	VolSearchEntry buffer[VolSearchBufferSize];
+	// buffer size must include an extra entry for an end of list marker 
+	VolSearchEntry buffer[MaxVolumeCount + 1];
```
DanRStevens: Comment capitalization is different here. Should we call the last entry a "null entry" or a "terminator entry"?

DanRStevens:
Your idea at the end of the first post sounded reasonable. Though yeah, you can get by without it.

VolList.h
Brett208: Thanks for fixing the punctuation inconsistency. I think you are correct to say terminator as it explains better what is going on than saying null.

TestModule/TestFunctions.cpp
Brett208: Now that the function's goal has been generalized to testing volumes, a little better documentation on what is happening wouldn't hurt.

Brett208:
Thanks for reviewing. Merging now.

PR #24, opened 2018-12-30
Eliminate cast by replacing with nullptr

DanRStevens:


Brett208:
Compiled in debug mode and it ran fine by loading TestModule using op2ext. Good change for readability as well as removing the cast.

PR #26, opened 2018-12-30
Vol list update

DanRStevens:
Transform `volPaths` from a `vector<vector<char>>` to `vector<string>`.

### DIFF
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -14,14 +14,14 @@ void VolList::AddVolFile(std::string volPath)
 // Note: Addresses of the original array (at various offsets) are hardcoded into several instructions
 void VolList::LoadVolFiles()
 {
-	std::size_t volEntryListSize = CreateVolSearchEntryList();
+	CreateVolSearchEntryList();
 
 	// Addresses at the start of the array are used for loop initial conditions
 	auto* arrayStart1 = &volSearchEntryList[0];
 	auto* arrayStart2 = &volSearchEntryList[0].unknown1;
 	// Addresses at the end of the array are used for loop termination conditions
-	auto* arrayEnd1 = &volSearchEntryList[volEntryListSize - 1];
-	auto* arrayEnd2 = &volSearchEntryList[volEntryListSize].unknown1;
+	auto* arrayEnd1 = &volSearchEntryList[volSearchEntryList.size() - 1];
+	auto* arrayEnd2 = &volSearchEntryList[volSearchEntryList.size() - 1].unknown1;
 
 	// Patch instructions so hardcoded references to old array now point to new array
 
@@ -52,19 +52,14 @@ void VolList::LoadVolFiles()
 // Changes to strings (such as by a vector re-allocation which moves strings), will invalidate cached c_str() pointers.
 // In particular, MSVC and other compilers implement the Small String Optimization (SSO).
 // SSO will exhibit problems for small strings when improperly cached c_str() pointers are invalidated by a move.
-std::size_t VolList::CreateVolSearchEntryList()
+void VolList::CreateVolSearchEntryList()
 {
-	// Buffer must include an extra terminator entry for an end of list marker
-	volSearchEntryList = std::make_unique<VolSearchEntry[]>(volPaths.size() + 1);
+	volSearchEntryList.clear();
 
-	std::size_t volEntryListSize = 0;
-	for (volEntryListSize; volEntryListSize < volPaths.size(); ++volEntryListSize) {
-		volSearchEntryList[volEntryListSize] = VolSearchEntry{ volPaths[volEntryListSize].c_str(), 0, 1, 0 };
+	for (const auto& volPath : volPaths) {
+		volSearchEntryList.push_back(VolSearchEntry{ volPath.c_str(), 0, 1, 0 });
 	}
 
-	// Add end of volFileEntries search item.
-	volSearchEntryList[volEntryListSize] = VolSearchEntry{ nullptr, 0, 1, 0 };
-	volEntryListSize++;
-
-	return volEntryListSize;
+	// Add end of volFileEntries search item (terminator entry)
+	volSearchEntryList.push_back(VolSearchEntry{ nullptr, 0, 1, 0 });
 }
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -29,7 +29,7 @@ void VolList::LoadVolFiles()
 
 	Op2MemSetDword((void*)0x00471070, arrayStart1);
 
-	//memcpy((void*)0x00471142, &arrayStart2, 4);
+	Op2MemSetDword((void*)0x00471142, arrayStart2);
 	Op2MemSetDword((void*)0x004711DA, arrayStart2);
 	Op2MemSetDword((void*)0x00471206, arrayStart2);
 	Op2MemSetDword((void*)0x0047136C, arrayStart2);
@@ -38,6 +38,7 @@ void VolList::LoadVolFiles()
 	Op2MemSetDword((void*)0x00471439, arrayStart2);
 	Op2MemSetDword((void*)0x00471474, arrayStart2);
 
+	Op2MemSetDword((void*)0x0047115E, arrayEnd1);
 	Op2MemSetDword((void*)0x0047126E, arrayEnd1);
 	Op2MemSetDword((void*)0x0047128B, arrayEnd1);
 	Op2MemSetDword((void*)0x00471389, arrayEnd1);
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -57,9 +57,9 @@ void VolList::CreateVolSearchEntryList()
 	volSearchEntryList.clear();
 
 	for (const auto& volPath : volPaths) {
-		volSearchEntryList.push_back(VolSearchEntry{ volPath.c_str(), 0, 1, 0 });
+		volSearchEntryList.push_back(VolSearchEntry::Init(volPath.c_str()));
 	}
 
 	// Add end of volFileEntries search item (terminator entry)
-	volSearchEntryList.push_back(VolSearchEntry{ nullptr, 0, 1, 0 });
+	volSearchEntryList.push_back(VolSearchEntry::Init(nullptr));
 }
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -18,10 +18,10 @@ void VolList::LoadVolFiles()
 {
 	std::size_t volEntryListSize = CreateVolSearchEntryList();
 
-	VolSearchEntry *vol = &volSearchEntryList[0];
-	int *vol2 = &volSearchEntryList[0].unknown1;
-	int *vol3 = &volSearchEntryList[volEntryListSize].unknown1;
-	VolSearchEntry *vol4 = &volSearchEntryList[volEntryListSize - 1];
+	auto* vol = &volSearchEntryList[0];
+	auto* vol2 = &volSearchEntryList[0].unknown1;
+	auto* vol3 = &volSearchEntryList[volEntryListSize].unknown1;
+	auto* vol4 = &volSearchEntryList[volEntryListSize - 1];
 
 	// Patch instruction references to old array
 
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -20,8 +20,8 @@ void VolList::LoadVolFiles()
 	auto* arrayStart1 = &volSearchEntryList[0];
 	auto* arrayStart2 = &volSearchEntryList[0].unknown1;
 	// Addresses at the end of the array are used for loop termination conditions
-	auto* arrayEnd1 = &volSearchEntryList[volSearchEntryList.size() - 1];
-	auto* arrayEnd2 = &volSearchEntryList[volSearchEntryList.size() - 1].unknown1;
+	auto* arrayEnd1 = &volSearchEntryList[volSearchEntryList.size()];
+	auto* arrayEnd2 = &volSearchEntryList[volSearchEntryList.size()].unknown1;
 
 	// Patch instructions so hardcoded references to old array now point to new array
 
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -3,7 +3,6 @@
 #include "GlobalDefines.h"
 #include <utility>
 
-VolList::~VolList() { }
 
 void VolList::AddVolFile(std::string volPath)
 {
--- a/VolList.cpp
+++ b/VolList.cpp
@@ -25,7 +25,7 @@ void VolList::LoadVolFiles()
 	auto* arrayEnd1 = &volSearchEntryList[volEntryListSize].unknown1;
 	auto* arrayEnd2 = &volSearchEntryList[volEntryListSize - 1];
 
-	// Patch instruction references to old array
+	// Patch instructions so hardcoded references to old array now point to new array
 
 	Op2MemSetDword((void*)0x00471070, arrayStart1);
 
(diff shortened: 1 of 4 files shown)


Brett208:
When running the test module and implementing this code, a half dozen dialog boxes similar to the one below appear before Outpost 2 exits (crashes not sure?). Something about switching from a `vector<vector<char>>` to `vector<string>` causes this. I fought with this for a while back when first updating op2ext to C++11. I think there is some history in the forums if you care to dig it up. 

I never did figure out what the real error was. I suspected it was associated with the contents of char[] becoming garbled perhaps by the string going out of scope and deleting itself but before a volume is loaded. Purely speculation though.

error modal dialog vol file load failure

-Brett

DanRStevens:
I think I see the problem. We use [`.c_str()`](https://en.cppreference.com/w/cpp/string/basic_string/c_str) to copy a pointer to the string. However, the lifetime of the returned pointer is not guaranteed. In particular, the documentation has the following note:

> The pointer obtained from c_str() may be invalidated by:
> - Passing a non-const reference to the string to any standard library function, or
> - Calling non-const member functions on the string, excluding operator[], at(), front(), back(), begin(), rbegin(), end() and rend().

As the strings are added to the vector, the vector may need to reallocate space and move the `std:string` objects to the new memory. The move operation requires calling non-const member functions. That could cause the string buffers to be re-allocated, and invalidate the memory previously returned by `.c_str()`.

In particular, if the `std::string` implementation uses the [small string optimization](https://blogs.msmvps.com/gdicanio/2016/11/17/the-small-string-optimization/), which MSVC does, then for sure those buffers will move and be invalidated when the `std::string` instances move. The short string optimization stores the contents of short strings directly in the `std::string` object itself, rather than in an external buffer pointed to and managed by fields internal to the `std::string` object. Less indirection, and no need for a separate call to `new` for memory allocation, though it has a fixed upper bound on the string length for which this can be applied. Beyond that, it must allocate an external buffer and manage access to it. As most of the filename strings here are quite short, they likely all fit within the small string optimization limit.

Brett208:
Quality research as usual. I had no idea the compiler could optimize std::string like that. I'm following the recommendation to wait until after LoadVolFiles is called to prepare volSearchEntryList. Good idea that will solve the issue. May be worth adding a comment about avoiding SSO there to make sure we don't go down the rabbit hole again.

I don't understand how this lifts the max volume count from being 31. Actually I have no idea how the code injects volumes into Outpost 2 in the first place. The comments on that part are poorly written coupled with me not understanding the general concepts. I am ok taking your word for it though.

I suppose we can just update the test to stress test adding a large number of volumes without breakfast ng Outpost 2 instead of a hard max.

DanRStevens:
The vector is not size limited. Only the static buffer is. The static buffer is an implementation detail. It's a holdover from old code, when it was done simply to avoid manual memory management. We weren't expecting many VOL files, nor much in the way of custom VOL files, so it seemed like a reasonable assumption.

Since we're already going with dynamic memory allocation for the vector, we might as well just go all the way.

When VOL files are loaded, the `volPaths` vector becomes fixed. We no longer allow additions to it. At this point, we know exactly how much memory we need for the `VolSearchEntry` array, so we can just dynamically allocate it at that time, rather than using a fixed compile time sized buffer.

The `VolSearchEntry` array gets loaded into Outpost 2 after it's contents become read-only. Outpost2.exe had its own internal `VolSearchEntry` array. The `LoadVolFiles()` method updates Outpost2.exe code to replace those references to its internal buffer and point them at our buffer instead.

DanRStevens:
I've moved the non-bug related commits to a new pull request (#28). Hopefully they can be fast tracked. I haven't removed those commits from this pull request at this time. I'll take care of that later, probably with a rebase/force update.

Brett208:
Took a quick stab at trying to defer creation of volSearchEntryList. Unfortunately receive an exception from a memory patcher later. Feel free to sift through or ignore and solve on you own.

New Header and CPP file based off this branch:
[VolList.zip](https://github.com/OutpostUniverse/op2ext/files/2719441/VolList.zip)

-Brett

**Exception**
Exception thrown at 0x250065F3 (NetHelper.dll) in Outpost2.exe: 0xC0000005: Access violation reading location 0x00000198. occurred

**Call Stack**
NetHelper.dll!std::_Hash<std::_Umap_traits<unsigned char *,unsigned char,std::_Uhash_compare<unsigned char *,std::hash<unsigned char *>,std::equal_to<unsigned char *> >,std::allocator<std::pair<unsigned char * const,unsigned char> >,0> >::lower_bound(unsigned char * const & _Keyval) Line 742	C++
NetHelper.dll!std::unordered_map<unsigned char *,unsigned char,std::hash<unsigned char *>,std::equal_to<unsigned char *>,std::allocator<std::pair<unsigned char * const,unsigned char> > >::operator[](unsigned char * const & _Keyval) Line 310	C++
 NetHelper.dll!Patcher::MemPatch::Disable(bool force) Line 118	C++
NetHelper.dll!Patcher::MemPatch::~MemPatch() Line 76	C++
[External Code]	
[Frames below may be incorrect and/or missing, no symbols loaded for ntdll.dll]	Unknown

DanRStevens:
That looks like a reasonable first stab at it. I think you might have an issue with the offset in:
```cpp
	int *vol3 = &volSearchEntryList[volPaths.size() - 1].unknown1;
	VolSearchEntry *vol4 = &volSearchEntryList[volPaths.size() - 2];
```

I think you might be correcting in the wrong direction. You need to include the terminator entry, which is not counted in `volPaths.size()`.

DanRStevens:
Side note: I assume the zip was because you didn't want to commit broken changes on top of this branch. If that's the case, you could potentially have created a new branch off of this branch as the starting point. That new branch could then have been pushed up to GitHub to share the changes.

Fixed code could then be merged back here. (With or without a merge commit).

DanRStevens:
Ok, I did the rebase and force push I'd suggested earlier. This cut out all the changes that have already been merged to master. That should keep this thread more focused on the changes related to the bug. It's also based on top of the other changes, which makes it easier for me to compile and test changes with MinGW.

DanRStevens:
I copied your changes on top of this branch, and then pushed them as a new branch: volListUpdateDeferCreation

Brett208:
Besides being broken I didn't want to make your force push more difficult by adding things that broke the code before the force push was made. 

Thanks for including my changes in the separate branch. I should have time later today to poke at it some more. Will check online first to see if you push any more changes before working.

So, what is the purpose of this branch now? To remove the workaround in the VolList constructor and allow loading an almost infinite number of vol files into Outpost 2? (I'm happy with these goals, just asking)

Thanks,
-Brett

DanRStevens:
Originally the change was to replace the weird looking `vector<vector<char>>` with `vector<string>`. Though that introduced a bug due to incorrect usage of `.c_str()`. The original structure meant the pointers returned by `.c_str()` could be invalidated before they were used. The plan now is to restructure the code to delay building the `VolSearchEntry` array to allow for correct usage of `.c_str()`. That would allow for the original change of switching to `vector<string>`. Additionally, such restructuring would also support uncapping the limit on the number of loaded VOL files. If the `VolSearchEntry` array is constructed later, it can also be allocated later. That allows precise sizing, rather than trying to guess a suitable upper bound for a fixed size buffer.

Actually, come to think of it, we can probably just use a `vector<VolSearchEntry>`. With the newer C++ standards, I believe we get guarantees about the internal structure of the vector (contiguous memory, and hopefully no padding).

----

I see it looks like you got things working in the other branch. I'll try to review shortly.

DanRStevens:
Ok, I made some updates to the patching code, which I think improves readability, and fixes some bit rot, though the changes will need some testing.

VolList.cpp
```diff
@@ -1,80 +1,70 @@
...
-	VolSearchEntry *vol = &volSearchEntryList[0];
-	int *vol2 = &volSearchEntryList[0].unknown1;
-	int *vol3 = &volSearchEntryList[volFileCount].unknown1;
-	VolSearchEntry *vol4 = &volSearchEntryList[volFileCount - 1];
+	// Addresses at the start of the array are used for loop initial conditions
+	auto* arrayStart1 = &volSearchEntryList[0];
+	auto* arrayStart2 = &volSearchEntryList[0].unknown1;
+	// Addresses at the end of the array are used for loop termination conditions
+	auto* arrayEnd1 = &volSearchEntryList[volEntryListSize - 1];
+	auto* arrayEnd2 = &volSearchEntryList[volEntryListSize].unknown1;
```
Brett208: I suspect this is a bug. It is reading a value past the end of the container. Perhaps it should read:

```suggestion
 	auto* arrayEnd2 = &volSearchEntryList[volEntryListSize - 1].unknown1;
```

Brett208:
Thanks for cleaning up the comments and code in `LoadVolFiles`. It had been bugging me for about a year now.

I think that I uncovered an indexing bug in `LoadVolFiles`. See comment above. Interestingly enough it compiled and ran fine before changing. 

I changed out the unique_ptr<c-style-array> for a vector. That is what uncovered the issue by throwing an exception when attempting to access an index outside of the vector. 

Everything compiles and runs fine in both DEBUG and RELEASE builds with Outpost 2. Program exits with a code 0. Then again it was exiting with a code 0 beforehand as well so not sure that says too much.

Besides running the test module in debug and release mode and making sure I can start a game are there any other tests you want me to perform? This proves that volumes are being loaded properly in general I suppose.

Brett

VolList.cpp
DanRStevens: I believe it's supposed to be 1 past the end, much like how the loop termination condition is typically 1 past the end. I'll add more details a bit later.

DanRStevens: Actually, now that I'm looking at the assembly, there is a bug. Both end entries should use `volEntryListSize`. There should be no `- 1` on either of them.

DanRStevens: Explanation:

When you write a for loop to iterate over an array or vector, the structure is:
```cpp
for (size_t i = 0; i < someVector.size(); ++i) {
  // ... use someVector[i]
}
```

You'll notice the loop termination condition uses `i < someVector.size()`, even though `.size()` is 1 past the end. The indexes only go from `0`..`(someVector.size() - 1)`.

Now in the above example, the only use of `i` in the loop body, is to index into `someVector`, with no index adjustments, and no raw uses of `i` in a numeric expression. That is, this loop is an ideal candidate to convert to a range based for loop:
```cpp
for (auto& someElement : someVector) {
  // ... use someElement
}
```

In a range based for loop, there is no loop index. It could be implemented internally with a loop index, but there is no requirement to do so. In fact, there's a good chance it won't use a loop index. It needs the element address, not the index of it. As such, it's more direct to store the address of the element. Such a loop might be implemented similar to:
```cpp
for (auto* elementAddr = &someVector[0]; elementAddr < &someVector[someVector.size()]; elementAddr++) {
  // ... use *elementAddr
}
```

You'll notice the loop termination condition is now `elementAddr < &someVector[someVector.size()]`, even though `&someVector[someVector.size()]` is an address past the end of the vector.

Similarly, if the elements were structs, and the loop body was accessing a single struct field:
```cpp
for (size_t i = 0; i < someVector.size(); ++i) {
  // ... use someVector[i].fieldX
}
```

The optimized form would then be:
```cpp
for (auto* fieldXAddr = &someVector[0].fieldX; fieldXAddr < &someVector[someVector.size()].fieldX; fieldXAddr = reinterpret_cast<typeof(fieldXAddr)>(reinterpret_cast<char*>(fieldXAddr) + sizeof(someVector::value_type))) {
  // ... use *fieldXAddr
}
```

Again, note the end address is `&someVector[someVector.size()].fieldX`, which is past the end of the vector. In fact, with the offset of `fieldX` added to it, it could actually be quite a bit beyond the end of the vector.

Note the awful code in the 3rd position of the `for` loop above for incrementing the loop. Normally `pointer++` behaves internally as `pointerValue += sizeof(*pointer)`. That is, it increments by the size of the pointed to type. Similarly `pointer + someInt` behaves internally as `pointerValue + someInt * sizeof(*pointer)`. Here the array element is a struct, so you want to increment by the struct size. The struct size is not strongly related to the field size (what the pointer points to). It's at least as big as the field it contains, and possibly not a multiple of the field size. Hence to represent that increment accurately, and in general, you'll need some ugly casting to undo the internal treatment of pointer addition.

----

What this is all saying, is that once the code is compiled, you'll often have addresses encoded in the instruction stream, rather the indexes. For loop termination conditions, those addresses will be beyond the end of the array.

DanRStevens:
I'm curious about that exception you said was thrown. As far as I'm aware, `vector::operator[]` is unchecked, and does not throw exceptions for out of bounds access. Instead, you are in the realm of undefined behaviour if you try to dereference/access that element. Taking the address of it should be fine, which is commonly done with iterators.

I don't see any warnings or errors on AppVeyor, though it's not actually running the code.

Sounds like you tested as I expected. I assumed just basic run time testing of the release build, making sure you can start a game, and that it exits cleanly.

DanRStevens:
If this code works for you, and doesn't produce any warnings/errors/crashes, I think we may be ready to merge this.

Brett208:
I retested the code a little closer. Running a clean command and then compiling in RELEASE mode, everything works fine. Running a clean command and then compiling in DEBUG mode, when running I get the following modal dialog:

debugassertionfailure

When clicking ignore, I get the following: Unhandled exception at 0x0F28E906 (ucrtbased.dll) in Outpost2.exe: An invalid parameter was passed to a function that considers invalid parameters fatal.

If I then compile in RELEASE mode without cleaning, I get the same dialog and exception. Cleaning the project, then rebuilding in RELEASE mode fixes is. I presume this is just a red herring caused by the fact that VS notices that nothing has changed in the code base and isn't properly recompiling even though I'm switching modes.

I think MSVC is providing more exhaustive runtime error checking when compiled in DEBUG mode that notices we are calling for an index outside of the vector. I'm guessing our options are to disable this check or just switch back to using the unique pointer with c style array solution. I think maybe in this case I would prefer just switching the code back to the unique pointer solution. I haven't researched it yet, but if we can locally disable this runtime check somehow I could live with that. I'd probably not want to disable the check project/solution wide.

Thanks,
-Brett

DanRStevens:
For reference, I came across this:
[May I take the address of the one-past-the-end element of an array? ](https://stackoverflow.com/questions/3144904/may-i-take-the-address-of-the-one-past-the-end-element-of-an-array)
[Take the address of a one-past-the-end array element via subscript: legal by the C++ Standard or not?](https://stackoverflow.com/questions/988158/take-the-address-of-a-one-past-the-end-array-element-via-subscript-legal-by-the)

The debate seems to be about whether or not `&*p` is considered a no-op, or taking the reference of a de-reference. As in, `&someVector[someVector.size()]` is basically equivalent to `&(*(p + size))`. The end result clearly doesn't need to touch memory, but technically, as per the grammar and language rules, it may consider there to be a de-reference in there, and it is undefined behaviour to de-reference past the end of the array.

----

Hmm, maybe we can use an end iterator instead.

DanRStevens:
Technically this still takes the address of a struct de-reference operator, so I'm not sure if this really sidesteps the problem. I'd be surprised if the compiler complained about it though.

Brett208:
Retested and got the following runtime failure (slightly different than last time).

second debug assertion

Pressing ignore gives the following:

Unhandled exception at 0x0F8CE906 (ucrtbased.dll) in Outpost2.exe: An invalid parameter was passed to a function that considers invalid parameters fatal.

-Brett

DanRStevens:
Fascinating. Ok, I stand corrected.

Maybe we should just add back the `-1` on the index for now. If we don't process the terminator entry, it should be fine. I don't image any code really relies on that entry in any significant way.

PR #29, opened 2019-01-01
Add Linux makefile

DanRStevens:
Ok, this isn't quite complete, but it helps me test, and I thought I'd share it now.

This makefile is able to compile all the .cpp files into .o files. Currently it fails during the link step. There are 2 undefined references that should come from the Outpost2DLL project, 1 undefined reference to `_ReturnAddress`, and 1 undefined reference for `WinMain`. Details:
```
i686-w64-mingw32-g++ .build/obj/Logger.o .build/obj/VolList.o .build/obj/op2ext-Internal.o .build/obj/OP2Memory.o .build/obj/GlobalDefines.o .build/obj/FileSystemHelper.o .build/obj/DllMain.o .build/obj/WindowsModule.o .build/obj/IpDropDown.o .build/obj/op2ext.o .build/obj/IniModuleLoader.o .build/obj/ConsoleModuleLoader.o -LSubmodules/Outpost2DLL/Lib/ -o op2ext.dll -lOutpost2DLL -lstdc++fs -lws2_32
.build/obj/DllMain.o: In function `@_Z7ExtInitP4TAppi@8':
/home/dstevens/code/opu/op2ext/DllMain.cpp:98: undefined reference to `_imp___ZN4TApp4InitEv'
.build/obj/DllMain.o: In function `@_Z11ExtShutDownP4TAppi@8':
/home/dstevens/code/opu/op2ext/DllMain.cpp:104: undefined reference to `_imp___ZN4TApp8ShutDownEv'
.build/obj/op2ext.o: In function `Log':
/home/dstevens/code/opu/op2ext/op2ext.cpp:107: undefined reference to `_ReturnAddress'
/usr/lib/gcc/i686-w64-mingw32/7.3-win32/../../../../i686-w64-mingw32/lib/../lib/libmingw32.a(lib32_libmingw32_a-crt0_c.o): In function `main':
./mingw-w64-crt/crt/crt0_c.c:18: undefined reference to `WinMain@16'
collect2: error: ld returned 1 exit status
makefile:29: recipe for target 'op2ext.dll' failed
make: *** [op2ext.dll] Error 1
```

This is however enough to get the code to run through the compiler and get diagnostic output. That alone could have some value for a CI build.

### DIFF
--- a/makefile
+++ b/makefile
@@ -30,6 +30,10 @@ SRCS := $(shell find $(SRCDIR) -maxdepth 1 -name '*.cpp')
 OBJS := $(patsubst $(SRCDIR)/%.cpp,$(OBJDIR)/%.o,$(SRCS))
 FOLDERS := $(sort $(dir $(SRCS)))
 
+# Default is to only compile, not link. MinGW is unable to link this project.
+# This is unlikely to be fixed anytime soon, so a default compile only step
+# can be used to scan the code for errors and warnings, without failing.
+default: $(OBJS)
 all: $(OUTPUT)
 
 $(OUTPUT): $(OBJS)


DanRStevens:
A couple of reference links concerning `_ReturnAddress` with MinGW that might be relevant:
[MinGW Compilation error: undefined reference to `_ReturnAddress'](https://bugzilla.mozilla.org/show_bug.cgi?id=1432790)
[_ReturnAddress C-RunTime source code](https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-crt/crt/gs_support.c)

Looks like the CRT converts the `_ReturnAddress` intrinsic to a GCC compiler builtin.

Brett208:
If this is successful, would you be able to compile op2ext on Linux using MinGW and then distribute it with new releases of Outpost 2? I have been happy with the MSVC results although it would be nice if you could also compile and release in case I'm not around. 

Or are we just looking to git the minGW compilation warnings to improve the code base and allow you to compile on Linux for error checking?

Thanks,
Brett

DanRStevens:
Mostly for ease of development on my end, though also for the additional error checking.

Previously:
(Check history to find my impossible to remember and type build command)
`mingw -std=c++17 -DOP2EXT_INTERNAL_BUILD *.cpp -LSubmodules/Outpost2DLL/Lib/ -lOutpost2DLL -lws2_32 -lstdc++fs`
(Side note: this is using the alias `alias mingw='i686-w64-mingw32-g++'`)
(Wait about 1 minute)
(Fix stupid syntax errors)

Now:
`make`
(wait a few seconds, fix errors)

Not only is the build command easier to remember and type, but it also uses .o files so it only needs to recompile the source files that have changed, making it much much faster.

----

You're right, it would be possible to build from Linux and distribute with the releases.

In practice, I doubt I will distribute a MinGW build with the release. Part of that is the binary releases are tracked in SVN, which stores everything as diffs. As the two compilers are likely to produce very different output, that would probably help balloon the size of the SVN repository. It could be especially bad if we were switching back and forth between compilers for each release.

Technically, even sticking with MSVC, a small compiler upgrade, or a small project upgrade could both cause massive changes in the compiled output, which would result in a bigger SVN diff. In practice however, this doesn't tend to happen within a compiler tool chain.

DanRStevens:
After playing with the makefile a bit, I think some of the link errors are fairly fundamental. There is a name mangling difference between MSVC and MinGW. That name mangling difference is related to a lack of a standard ABI for C++ code. Even if imports were somehow renamed, there's a good chance they still wouldn't link correctly, and would lead to runtime errors. It seems unlikely we would get a good working build from MinGW, at least not without modifying the `TApp` related code (possibly by using inline assembly).

As the makefile is still useful for testing changes on Linux, and for getting compiler feedback on syntax errors or warnings (some of which might not be caught by MSVC), I think this still has value and should be merged. As such, I've added a usability note about the problems, which are unlikely to be fixed.

Brett208:
Good to know about the name mangling being different between compilers. Hopefully this will help you some on Linux.

-Brett

Issue #31, opened 2019-01-08
MinGW Warnings


DanRStevens:
I noticed a few warnings from MinGW from my attempts at compiling on Linux.

```
FileSystemHelper.cpp: In function ‘std::__cxx11::string GetPrivateProfileStdString(std::__cxx11::string, std::__cxx11::string, std::__cxx11::string)’:
FileSystemHelper.cpp:37:7: warning: unused variable ‘returnSize’ [-Wunused-variable]
   int returnSize = GetPrivateProfileString(sectionName.c_str(), key.c_str(), "", buffer, currentBufferSize, filename.c_str());
       ^~~~~~~~~~
```

```
ConsoleModuleLoader.cpp: In member function ‘void ConsoleModuleLoader::LoadModule()’:
ConsoleModuleLoader.cpp:65:50: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (GetFileAttributesA(moduleDirectory.c_str()) == -1) {
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

```
ConsoleModuleLoader.cpp: In member function ‘std::__cxx11::string ConsoleModuleLoader::ParseLoadModCommand(std::vector<std::__cxx11::basic_string<char> >)’:
ConsoleModuleLoader.cpp:143:48: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (GetFileAttributesA(modDirectory.c_str()) == -1) {
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

```
op2ext.cpp:18:27: warning: ‘StubExt’ initialized and declared ‘extern’
 extern "C" OP2EXT_API int StubExt = 0;
                           ^~~~~~~
```

I came across a rather strong opinion about the last one, suggesting that warning should be considered a compiler bug. The other ones look quite legitimate though.

----

For reference:
[GetPrivateProfileString](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-getprivateprofilestring)
[GetFileAttributesA](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfileattributesa)

[...]

PR #33, opened 2019-01-08
Use returned size instead of `std::strlen`

DanRStevens:
This should be more efficient, and addresses the MinGW warning about the unused `returnSize` variable.

This partially addresses #31.

### DIFF
--- a/FileSystemHelper.cpp
+++ b/FileSystemHelper.cpp
@@ -2,7 +2,6 @@
 
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
-#include <cstring>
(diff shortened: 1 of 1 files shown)


DanRStevens:
I'm also wondering if we should use a `vector` or `string` data type here, rather than manual memory management. Maybe something for another branch.

Spurious build failure. It couldn't download form GitHub. I have no access to re-run the failed job.

DanRStevens:
Ok, merging as is then. If we want, we can always change out the manual memory management at a later point in time. I'm not going to create an issue for it though, too low priority.

PR #34, opened 2019-01-08
Use two stage extern declaration and init

DanRStevens:
Doing both at the same time results in a warning from MinGW.

This partially addresses issue #31.


DanRStevens:
Yeah, I think I prefer the shorter syntax too. Though I suppose this is more like the usual way of declaring a shared constant in a header files, and then defining it in a single implementation file. Just we don't need to share this constant with other compilation units, so no need to put it in a header file where it just pollutes the namespace. The export is only really used as a dummy entry to link to so that it causes the module to load.

PR #38, opened 2019-03-06
Create Log File Using Game Directory

### LINKED ISSUE #36: When debugger attached, log file is not created at root Outpost2 directory
When debugging with Visual Studio, the log file will no longer be created in the same directory as Outpost 2, but instead in the directory where the IDE project file is located.

I think proper behaviour would be to still create the log file in the Outpost 2 directory? This is the same issue that NetFix has when it creates its log file.

Brett208:
Resolves #36

### DIFF
--- a/Logger.cpp
+++ b/Logger.cpp
@@ -1,5 +1,6 @@
 #include "Logger.h"
 
+#include "FileSystemHelper.h"
 #include "GlobalDefines.h"
 #include <cstddef>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


Brett208:
The project uses \\ elsewhere already, so it keeps the path separator consistent. I think we are using some windows specific functions that return results with \\ so it is easiest way to integrate. I'm not opposed to switching, but someone should investigate our windows api calls first and figure out a strategy (if needed) to translate first.

DanRStevens:
Let's leave it as is.

As far as the Windows API is concerned, it will accept either "/" or "\" as a path separator, including a mix of the two within the same path. There's no need to translate. It's only the command shell (cmd.exe) that needs backslashes. Part of that is that many programs use the forward slash to denote command options. Example: `DIR /S` (Displays files recursively, traversing any subdirectories). As opposed to `DIR \S`, which looks for a folder names "S" off the root directory.

At any rate, let's just stick to being consistent, which for this project is "\".

PR #40, opened 2019-03-10
Add error details if module load fails

Brett208:
Decided to spend some time working with the Win32 API today. Hopefully the results are okay.


Brett208:
Cool. Thanks for looking over. Tested by deleting part of the contents of a console module dll btw.

PR #46, opened 2019-03-19
Copy string to buffer

### LINKED ISSUE #43: strncopy_s cannot take 0 for bufferSize
According to Microsoft's documentation, passing a destSize must be greater than 0. https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s?view=vs-2017

In order to support a paradigm of passing a destSize of 0 to get the actual required buffer size, I recommend we update to the function below. (Add a check for 0 up front).

```C++
size_t CopyStdStringIntoCharBuffer(const std::string& stringToCopy, char* buffer, size_t bufferSize)
{
	// Buffer must be one character larger than std::string since std::string does not include the null terminator.
	if (bufferSize == 0) {
	}

	strcpy_s(buffer, bufferSize, stringToCopy.c_str());

	if (bufferSize > stringToCopy.size()) {
		return 0;
	}

}
```

DanRStevens:
This closes #43.


DanRStevens:
We should look at adding unit tests to this project. I can think of a few test cases I'd like to run this function through.

DanRStevens:
I merged the branch from PR #53 to resolve the merge conflicts, and make integration easier.

Tests run from Linux. @Brett208, assuming we merge the other branch, can you add the new test file to the test project?

Brett208:
The function `CopyStdStringIntoCharBuffer` has a prototype only in its test file which, MSVC seems to be choking on. Guess maybe a compiler difference between MSVC and mingw? I'm finding the syntax of just forward declaring the function without it having a source header anywhere a bit confusing.

Brett208:
Dang, still broken. I think I like having the function moved into StringConversion with a formal prototype in StringConversion's header. Guess we could switch back to a forward declare in the test source file?

@DanRStevens, what do you think of this change (besides being broken)?

test/op2ext.test.cpp
```diff
@@ -1,9 +1,7 @@
+#include "../src/StringConversion.h"
```
DanRStevens: We should configure the `AdditionalIncludeDirectories` setting of the test project so we don't need relative references like this. Ideally the includes (and other code) in the test project should be good example code for client libraries to steal and model proper library usage from.

test/test.vcxproj
```diff
@@ -25,6 +25,7 @@
   <ImportGroup Label="PropertySheets" />
   <PropertyGroup Label="UserMacros" />
   <ItemGroup>
+    <ClCompile Include="op2ext.test.cpp" />
```
DanRStevens: I noticed there is no corresponding `test.vcxproj.filters` update for this line.

Issue #47, opened 2019-03-19
Add gtest test project


DanRStevens:
We should add a Google Test test project to this repository. Many of the helper methods could use some unit tests.

----

As this is a Windows only project, we'd need to use Mingw to compile from Linux. This requires a bit of special setup for Google Test. In particular, when building Google Test, we must set the command line option to cmake: `-DCMAKE_SYSTEM_NAME="Windows"`. Additionally there are some pthread issues when compiling gtest with Mingw. These are avoided by running cmake with the command line option `-Dgtest_disable_pthreads=ON`. These options can be added to the makefile.

There are additional complications when compiling the test code (as opposed to Google Test itself). To reference the Google Test library, we use `#include <gtest/gtest.h>`. The system installed headers are found in `/usr/include/gtest/`, hence we need an include directory of `-I/usr/include`. Unfortunately, this brings in standard Linux system header files, which conflict with the Mingw specific header files for Windows cross compilation. To get around this, we can change the include directory to be more specific with `-I/usr/include/gtest/`, and then update the source code inclusion to `#include <gtest.h>`. Unfortunately, Google Test itself includes it's other header files by using `gtest/` as part of the path, so those headers won't be found. The solution I found was to copy the `/usr/include/gtest/` folder to another location, perhaps `.build/include/gtest/`, and then add `-I.build/include/` as an additional include folder. That way it brings in gtest, but not the rest of the Linux system libraries.

This would be easier if Google Test didn't depend on the parent folder of the `gtest/` folder being in the include path.

Brett208:
I'll take care of adding a VS gtest project and moving the code around. 

Understand if you want mingw working for Linux tests and to be more productive. I would say not really required since we only plan to compile with MSVC for this project.

-Brett

DanRStevens:
Tagging:
Windows project structure updates:
PR #51 
PR #53 
Linux makefile updates:
PR #48 
PR #52 
Other:
PR #46 

----

Ok, in hindsight, I think we followed kind of a bad procedure on this one. It's resulted in a bit of a twisted interdependent mess. We had a large organizational change to add the test project while another branch was open. That creates merge conflicts. I was originally going to ask to merge the other branch first, but figured since the work here was already done, I'd just try and merge it. Additionally, as I wanted to add tests for the other branch, we needed a test project to run them from. Hence the mess of interrelated conflicting updates. As we've run into an issue with the test project setup, now multiple branches are kind of messed up.

[...]

PR #48, opened 2019-03-19
Update makefile

DanRStevens:
Cleans up the Linux makefile. This solves many of the Linux issues mentioned in Issue #47 for compiling and using Google Test with MinGW. (No attempt to address the Windows side of that issue, or adding any test project or test code).

----

Some object files were filtered from inclusion with test code due to references to Outpost2DLL or `_ReturnAddress`, which currently cause link errors on LInux with Mingw. In the case of Outpost2DLL, the problem is name mangling differences. I'm uncertain of the problem with `_ReturnAddress`.

Not addresses, are installing any needed Mingw specific libraries into Wine to run the test executable.

----

Linux only changes.

### DIFF
--- a/makefile
+++ b/makefile
@@ -76,6 +76,10 @@ gtest:
 	cp -r $(GTESTINCDIR) $(GTESTLOCALINCDIR)
 
 
+# Objects with references to Outpost2DLL or _ReturnAddress are a problem for the linker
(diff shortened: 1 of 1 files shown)


DanRStevens:
Tagging PR #50 for the `_ReturnAddress` fix.

PR #49, opened 2019-03-19
Update the test module's post build event to use environment variable

Brett208:
echo f | xcopy was required to tell xcopy that both copies were for files and not directories


Brett208:
Removing the filename allowed xcopy to work without needing to ask if the copy was a (f)ilename or a (d)irectory. So that was nice. Going to go ahead and merge to help test PR #50

PR #51, opened 2019-03-21
Create test project

Brett208:
Didn't make changes to makefile.

Should look through TestModule and move some tests over to unit tests as appropriate. I don't think we will be able to remove TestModule if we want to produce a DLL for testing op2ext's external hooks, but open to the idea if it exists.


IpDropDown.h
```diff
@@ -1,3 +0,0 @@
-#pragma once
-
```
Brett208: Looks like this file did not pickup on the rename. Not sure it matters being 3 lines long.

TestModule/CompatibilityTest.c
```diff
@@ -1,4 +1,4 @@
 // Tests if the external inteface has basic compatibility with C compiler.
```
Brett208: File rename not recognized on GitHub.

Brett208:
Is it still worth reviewing source code files so they contain an empty line at the end of each file. It doesn't do anything for me except keep GitHub from putting red Xs at the bottom of source code files. I understand some Unix based tools might require it? If there is a chance someone would want to use these tools on this project, I'm okay continuing, just want to make sure I'm not wasting time tracking them down and committing them.

-Brett

TestModule/CompatibilityTest.c
```diff
@@ -1,4 +1,4 @@
 // Tests if the external inteface has basic compatibility with C compiler. 
 // IE compiles without errors and warnings.
 
-#include "op2ext.h"
\ No newline at end of file
+#include "include/op2ext.h"
```
DanRStevens: Having the `include/` folder explicitly in the `#include` is probably not ideal. I think usually an include folder setting would be added to the project.

Brett208: I'm happy either way. 

-Brett

PR #54, opened 2019-03-23
Copy string to buffer - Try 2

### LINKED ISSUE #43: strncopy_s cannot take 0 for bufferSize

DanRStevens:
This closes #43. This is a rework of PR #46, which ran into some issues with a merge to add a MSVC project file for Google Test unit tests.

Current tests have been verified to pass on Linux.

### DIFF
--- a/StringConversion.cpp
+++ b/StringConversion.cpp
@@ -32,3 +32,18 @@ bool ConvertLPWToString(std::string& stringOut, const LPWSTR pw, UINT codepage)
 
 	return result;
 }
+
+std::size_t CopyStdStringIntoCharBuffer(const std::string& stringToCopy, char* buffer, std::size_t bufferSize)
+{
+	// Precheck valid pointer and non-zero buffer size to avoid wrap around or null termination problems
+	if (buffer != nullptr && bufferSize > 0) {
+		// Copy as much of the buffer as possible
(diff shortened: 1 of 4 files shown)


DanRStevens:
Ok, the new replacement method here uses well supported standard C++ library methods, rather than the old style C methods which were often either error prone or platform specific.

Based on Linux test results, and the discovery about the Windows test project setup problems in the other thread, I'm going to go ahead and merge this. That should help us unravel the twisted dependency issue concerning adding a test project. It also gives us some real test code to run when the test project is added.

PR #58, opened 2019-03-23
Add MSVC unit tests

Brett208:


Brett208:
Well, works now. Basically I added a new configuration to op2ext solution that builds it statically.

 - Add a static build configuration to op2ext's project and solution (DebugStatic)
 - Disable building TestModule in DebugStatic (Should only build dynamic (DLL))

Appveyor is attempting to run unit tests in all configurations right now. We need a configuration matrix or some other construct to only allow unit tests to run  on DebugStatic. Also had to hard code the folder containing test.exe. If this looks good we can add a ReleaseStatic configuration.

We could name the configurations ReleaseUnitTests instead of Static too.

-Brett

DanRStevens:
Hmm, the build matrix issue seems like a big negative for testing. Basically we wouldn't be running tests for the Debug and Release configurations. That should give us pause. By only running tests on a different configuration, we make it much easier for problems to slip through the cracks and end up being released.

After breaking from my programming session earlier today, I was considering just adding a static compile option for use with testing, as you've done here. Now that I see it though, I'm having strong doubts.

In terms of maintaining testability, I think we might need the core of op2ext to be a static library, and then have a wrapper dynamic link library project that pulls in code from the static library. A test project can then also pull in code from the static library. In fact, we'd need a static library to do unit testing, as otherwise we'd be limited to only testing the exported interface, which is a relatively small area to test.

An alternative might be to include the core code into the test project itself. Though that means it needs to be compiled twice if you build both projects.

I'm feeling somewhat dissatisfied with all the solutions I can think of right now. Perhaps I should do a bit of research on how other people perform unit testing of non-exported stuff from DLLs. A quick Google search brings up a few hits.

Brett208:
Odd that you don't like the solution. I had the opposite reaction and was glad we figured out a way to both test the guts of op2ext through unit tests and leave the TestModule in place to test the externally available interface.

I think instead of a build matrix we should be able to just use an IF statement to tell it to run unit tests only if the configuration is DebugStatic. I pushed a commit to test. It looks like the IF statement is processed properly, but it didn't recognize the the command to run test.exe.

If we are performing unit tests on the non exported members of op2ext, how would the tested code be treated differently if we statically compile a common library vs just statically compiling op2ext itself? 

I don't follow how creating a library project decreases the chances of problems slipping through the cracks vs adding a different configuration to use when unit testing. To truly represent op2ext in its final form, we would have to force all unit tests to call through the public interface, op2ext.h after op2ext.dll was compiled, which would be restrictive on the tests. I'm by no means an expert on the subject though.

Thanks,
Brett

DanRStevens:
Remember there's a reason we build both the Debug and Release configurations. Sometimes only 1 of them fails. Those project settings can affect compilation, which affects how code runs. That can easily lead to certain tests passing or failing, based on build settings. It's also quite easy to make a mistake with project configuration, or to only change some value on one configuration rather than all of them.

Whenever code is compiled with different settings, there is a potential for different behaviour. We should strive to test the code under normal build conditions. Adding an additional configuration just for testing goes against that. In a way, it invalidates the testing.

----

In terms of dynamic link libraries, versus static link libraries, there is a difference in visibility.

Dynamic link libraries are more restrictive, in that they only provide access to the publicly exported methods. The other code is still there, it's just not exported in a visible way, and so you have no way to make direct calls into the internals. That means no unit testing of those internals.

If you have a static library, everything is public within the static library by default. At least, it's public in the same way that methods from different compilation units can link to each other simply by providing an appropriate forward declaration, or header include. It's public unless it was explicitly put into an anonymous namespace to make it impossible to access. That's how object files work, and how they can be linked together. A static library is then just an archive of object files. It's basically the equivalent of taking all your object files and zipping them up (with compression disabled) into one file. This makes unit testing very easy, as you just add a dependency on the static library, include the header, and call the method of interest.

----

Static libraries are precompiled code that hasn't been fully linked yet. That means you can link it into a DLL, where it becomes part of the DLL, or you can link it into a test project, where it becomes part of the test project. Either way, it was only compiled once, and with consistent settings between the two uses.

What I'm curious about, is if MSVC outputs a static library file as well when building a DLL, or if it can be made to also output a static library. That would make things very easy for unit testing, while still using the DLL build settings in a single project.

If that's not possible, wrap a DLL project around a static library project.

PR #53, opened 2019-03-22
Create test project - Cleanup

DanRStevens:
This is a cleanup of PR #51.

This one was a little trickier to get right. I found I had to move the commit to add trailing newlines to files ahead of the commit that ran dos2unix. Failing to do so resulted in merge conflicts and unrecognized renames.

I updated the Visual Studio project settings to add `include` as one of the `AdditionalIncludeDirectories`. That minimizes file changes during the rename, and gets rid of the weird folder in `#include "include/..."`. As the main project file `op2ext.vcxproj` did not already include this, it was a manual text addition from me. @Brett208, you may want to check that Visual Studio accepts it, and that it regenerates the project file with the setting in the same position. **Edit:** I saw some AppVeyor build failures for the Release configuration, which were related to this setting. In the old branch, the Debug configuration had this setting added when source files were moved (not when the include folder was created). There was no corresponding addition to the Release configuration. When I added the setting (when the include folder was created), I didn't notice the pre-existing copy in Debug added in the later commit. It didn't generate a merge conflict due to the moved position. In addition to that, I made a path error on my first attempt. That should explain the Release build failures. It should be cleaned up now.

I updated history to include `makefile` changes for each step along the way. The makefile should be functional at each stage.

PR #60, opened 2019-03-30
Update gitignore

DanRStevens:
Useful changes from the branches that need rework. (#53)

@Brett208, I wanted to double check the .ini ignore setting with you. I assume it's not needed anymore?


Brett208:
Correct, the new code should prevent the .ini file from forming in the wrong place. Probably fine to remove from the .gitignore at this point.

PR #61, opened 2019-03-30
Remove obsolete preprocessor define

DanRStevens:
I believe that define predated the current use of `OP2EXT_INTERNAL_BUILD`.


Brett208:
I probably forgot to remove it when introducing the new preprocessor directive

PR #62, opened 2019-03-30
Split project into DLL and Static library

DanRStevens:
Ok, I think these are the changes we need to the project structure. I was able to cobble this together using the project file edits from the cleanup branch. I have not yet added in the test project. That will be later work based on this.

@Brett208, please let me know how well this project structure works for you. I see it passing on AppVeyor, but that doesn't mean there can't be any local Visual Studio issues.


DanRStevens:
Fixed some merge conflicts with another branch that was recently merged.

PR #66, opened 2019-04-02
Implement code reorganization in project filter files

Brett208:
Move resource file from static to dll project (File was already placed properly, but tracked incorrectly in project).

I should have caught these omissions before approving the original branch that separated into static and dll folders. Lazy job on my part.


Brett208:
force push to remove last commit.

PR #67, opened 2019-04-03
Set output filename to op2ext.dll

Brett208:
Now that it is built from op2extDLL project, the default filename is op2extDLL.dll and must be manually configured.

I missed this detail earlier because I didn't manually delete op2ext.dll from Outpost2Path's location. So both an op2ext.dll and a op2extDLL.dll existed. The older op2ext.dll must have initially worked okay for my test to make sure the game used the module loader. Sometime after the initial test, things went south. This also prevented the debugger from finding the correct file and symbols.

### DIFF
--- a/srcDLL/op2extDLL.vcxproj
+++ b/srcDLL/op2extDLL.vcxproj
@@ -43,9 +43,11 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <LinkIncremental>false</LinkIncremental>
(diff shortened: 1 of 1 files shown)


Brett208:
I'm a bit on the fence. If you rename, it would be directory op2ext/op2ext, which doesn't really make sense. But it would prevent having to set a custom name in the project files (although setting the name is a pretty simple step). 

I'm going to merge this now as a bug fix for the current master. I don't mind if you want to change up the directory name and remove the property now or later or never.

DanRStevens:
You could rename the project file without renaming the directory. Anyway, something we can revisit later if we want.

PR #70, opened 2019-04-19
Update test project

Brett208:
There is an intention to compile using mingw. Not sure if we can remove the experimental includes for filesystem and accomplish this. If not, it will be easy to add a preprocessor directive to detect.

The static project was already setup to use C++17. I noticed the newer version of the gtest nuget package was already in use.

This branch re-enables 4996 warnings on the test project sense the newer 1.8.1 nuget gtest package no longer produces them.

### DIFF
--- a/TestModule/CountFilesByTypeInDirectory.cpp
+++ b/TestModule/CountFilesByTypeInDirectory.cpp
@@ -1,9 +1,9 @@
 #include "CountFilesByTypeInDirectory.h"
 #include <string>
 #include <algorithm>
-#include <experimental/filesystem>
+#include <filesystem>
 
-namespace fs = std::experimental::filesystem;
+namespace fs = std::filesystem;
 
 void ToLowerInPlace(std::string& x) {
 	std::transform(x.begin(), x.end(), x.begin(), ::tolower);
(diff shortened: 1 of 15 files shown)


Brett208:
I went ahead and moved the include for filesystem into the FileSystemHelper header. Seemed the most logical existing location. Didn't want to go too far down a rabbit hole of filesystem code updates in this branch.

Just stole the implementation that @DanRStevens used in OP2Utility. Thank you for writing it up as it was on my TODO list here.

Brett208:
I would like to merge this branch to keep from getting too far into future work outside of master. If the filesystem include change is undesirable, will work a new branch to fix.

TestModule/CountFilesByTypeInDirectory.cpp
```diff
@@ -1,9 +1,8 @@
 #include "CountFilesByTypeInDirectory.h"
+#include "../srcStatic/FileSystemHelper.h"
```
DanRStevens: There should already be an `AdditionalIncludeDirectories` setting for `../srcStatic/`. We should use that over relative include paths between the projects.

Issue #69, opened 2019-04-16
Method to determine if module is loaded


DanRStevens:
Some levels may be dependent on a module being loaded, such as an alternate rule set for the game. Perhaps these levels could be given a method to determine if a given module is loaded, or to retrieve a list of loaded modules.

This is something Crow was recently looking for. Such functionality seems most related to op2ext, as it controls which modules get loaded.

Brett208:
This would be helpful for multitek2 as well. I guess the user would have to include op2ext in their mission project (or at least the public header) My first stab would be:

```C++
// Returns the number of loaded modules (.ini and console combined)
size_t GetLoadedModuleCount();


// Retrieves the module name at the specified index. Use function GetLoadedModuleCount to determine how many module names to check. 
// A console module's name is the name of the directory the console module is stored in.
// An ini module name is the module's [section name] within the ini file
// Returns 0 on success. Returns the required minimum size of the buffer on failure.
// If an index beyond the loaded module count is passed, returns 0 and clears the buffer.
OP2EXT_API size_t GetLoadedModuleName(char* buffer, size_t bufferSize);
```

DanRStevens:
That looks like it could work.

Should be generic enough to implement whatever methods are needed. Not sure if we want to go there, but we could potentially also have a method to query if a specific module is loaded. That may be what people really want, though the above implementation is more generic with a greater number of use cases.

Brett208:
In addition to the 2 functions above, we could add:

```C
 #include <stdbool.h>

// Performs a case insensitive search of loaded module names, returning true if found.
// A console module's name is the name of the directory the console module is stored in.
// An ini module name is the module's [section name] within the ini file.
OP2EXT_API bool IsModuleLoaded(char* moduleName);
```

DanRStevens:
What happens if an ini section name matches a console module's directory name? I don't believe we keep any kind of tracking information around about what is loaded, or how it was loaded.

From a caller perspective, do we need to differentiate between the module types? They are rather incompatible, so maybe we should. That might apply to both sets of methods.

Brett208:
The console module loader and .ini module loader do not talk. So you could load a .ini and console module with the exact same name. Practically speaking I would discourage this behaviour. Not sure it is worth preventing/fixing as it never occurred to me to try in practice, nor have I heard anyone attempting. We could add a check to see if the module is already loaded internal to op2ext to prevent. Although that would maybe be best as a separate issue from this.

[...]

PR #71, opened 2019-04-20
Access module names

Brett208:
Finished a sprint towards incorporating #69.

Next step is probably adding similar functions to ini module loader.

I was pleasantly surprised that I could execute some limited unit tests against the ConsoleModuleLoader.

### DIFF
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -39,8 +39,7 @@ std::string ConsoleModuleLoader::GetModuleName()
 
 bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
 {
-	ToLowerInPlace(moduleName);
-	return moduleName == GetModuleName();
+	return ToLowerInPlace(moduleName) == GetModuleName();
 }
 
 int __fastcall GetArtPath(void*, int, char*, char*, char *destBuffer, int bufferSize, char *defaultValue)
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -20,6 +20,13 @@ ConsoleModuleLoader::ConsoleModuleLoader()
 	moduleDirectory = FindModuleDirectory();
 }
 
+ConsoleModuleLoader::ConsoleModuleLoader(const std::string& testModuleDirectory)
+{
+	OutputDebugString("Console Module constructed in test mode.");
+
+	moduleDirectory = testModuleDirectory;
+}
+
 std::string ConsoleModuleLoader::GetModuleDirectory()
 {
 	return moduleDirectory;
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -30,6 +30,12 @@ std::string ConsoleModuleLoader::GetModuleName()
 	return ToLower(moduleDirectory);
 }
 
+bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
+{
+	ToLowerInPlace(moduleName);
+	return moduleName == GetModuleName();
+}
+
 int __fastcall GetArtPath(void*, int, char*, char*, char *destBuffer, int bufferSize, char *defaultValue)
 {
 	strcpy_s(destBuffer, bufferSize, moduleDirectory.c_str());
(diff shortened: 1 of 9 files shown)


Brett208:
This is based on PR #70. I haven't figured out how to not include the base branch's changes in the PR report.

srcStatic/ConsoleModuleLoader.h
```diff
@@ -6,10 +6,14 @@
 class ConsoleModuleLoader {
 public:
 	ConsoleModuleLoader();
+	// Test Constructor
+	ConsoleModuleLoader(const std::string& testModuleDirectory);
```
DanRStevens: There's something about the comment that makes me want to remove it. I can understand why it's there though, and it probably should be there. Perhaps this relates to my earlier comments about dependency injection, and wondering if the constructors should be reworked such that a comment wouldn't be of value. Anyway, food for thought. No changes needed.

srcStatic/StringConversion.cpp
```diff
@@ -47,3 +48,13 @@ std::size_t CopyStdStringIntoCharBuffer(const std::string& stringToCopy, char* b
 	// Unable to copy the whole string, so return needed buffer size
 	return stringToCopy.size() + 1;
 }
+
+void ToLowerInPlace(std::string& x) {
+	std::transform(x.begin(), x.end(), x.begin(), ::tolower);
+}
```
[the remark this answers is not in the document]
Brett208: Would this look like:

```C++
std::string& ToLowerInPlace(std::string& x) {
	std::transform(x.begin(), x.end(), x.begin(), ::tolower);

	return x;
}
```

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -20,11 +20,29 @@ ConsoleModuleLoader::ConsoleModuleLoader()
 	moduleDirectory = FindModuleDirectory();
 }
 
+ConsoleModuleLoader::ConsoleModuleLoader(const std::string& testModuleDirectory)
+{
+	OutputDebugString("Console Module constructed in test mode.");
+
+	moduleDirectory = testModuleDirectory;
+}
```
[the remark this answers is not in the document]
Brett208: I am a fan of Dependency Injection. Although in the past I've taken it much too far, passing in 6-8 parameters into a constructor that didn't really contribute to testability.

As you mentioned, I think for the long term health of ConsoleModuleLoader, we should consider passing in the command line arguments and eliminate the empty constructor. There are some design changes to the overall structure of op2ext that may be required. I think maybe deferring this work to a more dedicated branch would be a better idea.

srcStatic/ConsoleModuleLoader.h
Brett208: Once someone refactors ConsoleModuleLoader to only have 1 constructor, won't be an issue anymore. In its current state, should not be used outside test environment.

Brett208:
Thanks for the quick review. Generally agree with all the suggestions. Need a bit more stick and rudder on how to implement ToLowerInPlace suggestion. I don't typically return references from functions.

srcStatic/StringConversion.cpp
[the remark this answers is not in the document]
DanRStevens: Yes, exactly.

DanRStevens:
Your suggested implementation for returning a reference was exactly what was needed.

The one rule is to never return a reference to a local variable (or the address of a parameter, which is also part of the function's stack frame). In this case, the reference is passed in as a parameter value (not the address of the parameter). In particular, it was passed in by the caller, where you're returning it to. Naturally that address had better be valid for the caller, otherwise they shouldn't have been passing it to the function.

Brett208:
Newlines at end of file fixed in commit to update ToLowerInPlace. Thanks for the help.

test/StringConversion.test.cpp
```diff
@@ -11,11 +11,9 @@ TEST(StringConversion, ToLower)
 TEST(StringConversion, ToLowerInPlace)
 {
 	std::string str("TEST");
-	ToLowerInPlace(str);
-	EXPECT_EQ("test", str);
+	EXPECT_EQ("test", ToLowerInPlace(str));
```
DanRStevens: You may want to check both the return value and the original value is modified:
```cpp
EXPECT_EQ("test", ToLowerInPlace(str));
EXPECT_EQ("test", str);
```

Without both checks, the `ToLower` method could be used to satisfy the test, even though it offers different behavior from `ToLowerInPlace`.

test/StringConversion.test.cpp
```diff
@@ -11,11 +11,9 @@ TEST(StringConversion, ToLower)
...
 
 	// Test Empty string
 	str = "";
-	ASSERT_NO_THROW(ToLowerInPlace(str));
-	EXPECT_EQ("", str);
+	EXPECT_EQ("", ToLowerInPlace(str));
```
DanRStevens: Ditto.

PR #73, opened 2019-04-21
Access ini module name

Brett208:
Finished short sprint allowing retrieving on ini module name

Next step is either trying to provide a mockup .ini file for unit testing or hooking the function into op2ext.h for external calling.

### DIFF
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -1,5 +1,4 @@
 #include "op2ext.h"
-
 #include "StringConversion.h"
 #include "OP2Memory.h"
 #include "FileSystemHelper.h"
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -111,7 +111,7 @@ OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
 
 OP2EXT_API size_t GetLoadedModuleCount()
 {
-	const std::size_t consoleModuleCount = (consoleModLoader.IsModuleLoaded() ? 1 : 0);
+	const std::size_t consoleModuleCount = consoleModLoader.IsModuleLoaded() ? 1 : 0;
 
 	return iniModuleLoader.Count() + consoleModuleCount;
 }
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -103,3 +103,8 @@ OP2EXT_API bool IsConsoleModuleLoaded(const char* moduleName)
 {
 	return consoleModLoader.IsModuleLoaded(moduleName);
 }
+
+OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
+{
+	return iniModuleLoader.IsModuleLoaded(moduleName);
+}
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -108,3 +108,10 @@ OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
 {
 	return iniModuleLoader.IsModuleLoaded(moduleName);
 }
+
+OP2EXT_API size_t GetLoadedModuleCount()
+{
+	const std::size_t consoleModuleCount = (consoleModLoader.IsModuleLoaded() ? 1 : 0);
+
+	return iniModuleLoader.Count() + consoleModuleCount;
+}
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -111,7 +111,5 @@ OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
 
 OP2EXT_API size_t GetLoadedModuleCount()
 {
-	const std::size_t consoleModuleCount = consoleModLoader.IsModuleLoaded() ? 1 : 0;
-
-	return iniModuleLoader.Count() + consoleModuleCount;
+	return iniModuleLoader.Count() + consoleModLoader.Count();
 }
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -34,6 +34,11 @@ std::string ConsoleModuleLoader::GetModuleName()
 	return ToLower(moduleDirectory);
 }
 
+bool ConsoleModuleLoader::IsModuleLoaded()
+{
+	return moduleDirectory != "";
+}
+
 bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
 {
 	return ToLowerInPlace(moduleName) == GetModuleName();
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -34,14 +34,14 @@ std::string ConsoleModuleLoader::GetModuleName()
 	return ToLower(moduleDirectory);
 }
 
-bool ConsoleModuleLoader::IsModuleLoaded()
-{
-	return moduleDirectory != "";
+std::size_t ConsoleModuleLoader::Count()
+{	
+	return moduleDirectory != "" ? 1 : 0;
 }
 
 bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
 {
-	if (moduleDirectory == "") {
+	if (Count() == 0) {
 		return false;
 	}
 
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -35,7 +35,7 @@ std::string ConsoleModuleLoader::GetModuleName()
 }
 
 std::size_t ConsoleModuleLoader::Count()
-{	
+{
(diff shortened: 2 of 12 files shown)


srcStatic/IniModuleLoader.cpp
```diff
@@ -54,6 +54,15 @@ bool IniModuleLoader::UnloadModules()
 	return result;
 }
 
+std::string IniModuleLoader::GetModuleName(std::size_t index)
+{
+	if (index >= modules.size()) {
+		return "";
```
DanRStevens: Perhaps a `throw` would be more appropriate here. As per the `std::string` return value, this code is strictly C++ anyway. The translation to a return value for the C interface of the DLL could be done at a higher layer.

test/IniModuleLoader.test.cpp
```diff
@@ -0,0 +1,13 @@
+#include "IniModuleLoader.h"
+#include <gtest/gtest.h>
+
+TEST(IniModuleLoader, InitializeNoModules)
+{
+	IniModuleLoader iniModuleLoader;
+	EXPECT_NO_THROW(iniModuleLoader.LoadModules());
+	
+	EXPECT_EQ("", iniModuleLoader.GetModuleName(0));
+	EXPECT_EQ("", iniModuleLoader.GetModuleName(1));
```
DanRStevens: My previous suggestion would affect this test, so it would instead be expected to throw.

srcStatic/IniModuleLoader.h
```diff
@@ -26,13 +26,15 @@ class IniModuleLoader
 	{
 		HINSTANCE handle;
 		DestroyModFunc destroyModFunc;
+		// The module's name is the same as the source ini file SectionName
+		std::string name;
```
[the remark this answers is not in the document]
Brett208: I would like to document somewhere that the moduleName is the same as the sectionName. It could be documented as a comment in the function GetModuleName if you prefer.

srcStatic/IniModuleLoader.cpp
```diff
@@ -18,9 +18,10 @@ void IniModuleLoader::LoadModules()
 void IniModuleLoader::LoadModule(std::string sectionName)
 {
 	IniModuleEntry moduleEntry;
+	moduleEntry.name = sectionName;
 
 	try {
-		LoadModuleDll(moduleEntry, sectionName);
+		LoadModuleDll(moduleEntry);
```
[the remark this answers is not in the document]
Brett208: Agreed that the code could be improved. I propose passing the module's handle out of LoadModuleDll as the solution. This way LoadModuleDll doesn't need the IniModuleEntry struct at all. Will push in a commit. A little different then proposed but I think results will be similar.

srcStatic/IniModuleLoader.cpp
Brett208: I don't see much use for the throw here. I would just be catching and translating into an empty string and I don't have a use case for the exception yet. Seems like extra work for no reason. Don't see anything wrong with a throw/catch solution though. If it is important to you, let me know and I'll implement.

srcDLL/op2ext.cpp
```diff
@@ -104,3 +103,15 @@ OP2EXT_API bool IsConsoleModuleLoaded(const char* moduleName)
...
 }
+
+OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
+{
+	return iniModuleLoader.IsModuleLoaded(moduleName);
+}
+
+OP2EXT_API size_t GetLoadedModuleCount()
+{
+	const std::size_t consoleModuleCount = (consoleModLoader.IsModuleLoaded() ? 1 : 0);
```
Brett208: Was not sure if parenthesis are needed for ternary operator use here. I have tended to avoid the ternary operator in the past since it can be confusing. Let me know if it would be better to separate into an if statement.

test/ConsolueModuleLoader.test.cpp
```diff
@@ -10,6 +10,7 @@ TEST(ConsoleModuleLoader, NoModuleLoaded)
 	// Returns empty string if no module available
 	EXPECT_EQ("", consoleModLoader.GetModuleName());
 
+	EXPECT_FALSE(consoleModLoader.IsModuleLoaded());
 	EXPECT_TRUE(consoleModLoader.IsModuleLoaded(""));
```
Brett208: Should this return false instead? It is a corner case where no module is loaded (stored as an empty string). We could have an empty string always return as false. This makes sense to me because you couldn't contain the console module directories in a the directory "", as that would likely be the root directory?

Brett208:
I'm thinking of stopping adding new code to this branch besides what is needed for cleaning up the work so it can be merged and doing new field work or heavy lifting in a different branch.

srcStatic/IniModuleLoader.h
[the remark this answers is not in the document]
DanRStevens: Actually, yes, that should probably be documented for `GetModuleName`. That's the public interface where callers will care about what a `moduleName` actually is. In contrast, the `IniModuleEntry` is a private struct which the user shouldn't know or care about.

srcStatic/IniModuleLoader.cpp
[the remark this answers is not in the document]
DanRStevens: Good point, that is another alternative. I didn't check closely, and had assumed both `handle` and `destroyModFunc` were set in that method, but they aren't. I suppose you could go either way. It may make sense to move setting the `destroyModFunc` field into that method.

The same could also be done for `initModFunc`. I considered storing that value too when I first wrote it. I didn't though, since there was no reason to store it after calling it. Might be cleaner code to store it in the struct though.

srcStatic/IniModuleLoader.cpp
DanRStevens: Sure, there's not much use, and that's largely because of the restrictions imposed by the C API, which is the only place that calls this code. Effectively it's pushing a C way of handling errors (with return codes) onto C++ code, that would normally use exceptions to report errors. It kind of feels like it pollutes the C++ code with old style C-isms.

I suppose you can apply the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) to justify keeping it as is. Not really a big deal I suppose.

srcDLL/op2ext.cpp
DanRStevens: You can eliminate the parenthesis. Ternary can be confusing to read, though I often prefer the more compact ternary form to multiline `if`/`else` blocks.

test/ConsolueModuleLoader.test.cpp
DanRStevens: I had wondered about this too. I don't have a clear answer. I suppose you could dump a module in the root game folder. Might even be useful for single file modules. It's either a single file or a single directory, so kind of the same either way. But then, there's an assumption about the name of the module DLL within the module folder, so that breaks things.

Either way you go, you'll need to document this corner case. Or maybe consider an alternate design that avoids this corner case.

srcDLL/op2ext.cpp
```diff
@@ -108,3 +108,10 @@ OP2EXT_API bool IsIniModuleLoaded(const char* moduleName)
 {
 	return iniModuleLoader.IsModuleLoaded(moduleName);
 }
+
+OP2EXT_API size_t GetLoadedModuleCount()
+{
+	const std::size_t consoleModuleCount = (consoleModLoader.IsModuleLoaded() ? 1 : 0);
+
+	return iniModuleLoader.Count() + consoleModuleCount;
```
DanRStevens: Maybe the ternary operation should be moved into a method on `consoleModuleLoader` for consistency.

test/ConsolueModuleLoader.test.cpp
Brett208: OK, agree on returning false. I was leaning that way as well.

Thanks,
Brett

srcDLL/op2ext.cpp
Brett208: The console module loader doesn't really make sense returning a count because it can only have 1 module. However, I like that this will make the ini module and console module loaders appear identical in function setup. I suppose it allows easier expansion if we ever support multiple console modules.

Brett208:
@DanRStevens, if most recent changes look good, recommend merging and moving on to next portion.

Thanks,
-Brett

srcDLL/op2ext.cpp
DanRStevens: Yes, good point, having the console module loader only support 1 module seems more like a current implementation detail. It could potentially be upgraded to support multiple modules at some point in the future.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -41,6 +41,10 @@ bool ConsoleModuleLoader::IsModuleLoaded()
 
 bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
 {
+	if (moduleDirectory == "") {
+		return false;
+	}
```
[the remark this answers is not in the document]
Brett208: Well, verified my decision to make the change anyways. :)

PR #77, opened 2019-04-24
Finalize module presence public interface

Brett208:
Not sure if it is too confusing how GetModuleName works (indexes ini modules first, then the console module).

Alternate implementation would be to have 2 functions, one that gets the name of the console module (if loaded), and one that gets the name of a ini module.

-Brett

### DIFF
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -99,6 +99,12 @@ OP2EXT_API void Log(const char* message)
 	logger.Log(message, FindModuleName(_ReturnAddress()));
 }
 
+
+OP2EXT_API bool IsModuleLoaded(const char* moduleName)
+{
+	return IsConsoleModuleLoaded(moduleName) || IsIniModuleLoaded(moduleName);
+}
+
 OP2EXT_API bool IsConsoleModuleLoaded(const char* moduleName)
 {
 	return consoleModLoader.IsModuleLoaded(moduleName);
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -119,3 +119,22 @@ OP2EXT_API size_t GetLoadedModuleCount()
 {
 	return iniModuleLoader.Count() + consoleModLoader.Count();
 }
+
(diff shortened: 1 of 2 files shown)


srcDLL/op2ext.cpp
```diff
@@ -119,3 +119,22 @@ OP2EXT_API size_t GetLoadedModuleCount()
...
+OP2EXT_API size_t GetLoadedModuleName(size_t moduleIndex, char* buffer, size_t bufferSize)
+{
+	// Ini modules are indexed first and console module is last
+
+	std::string moduleName;
+
+	if (moduleIndex < iniModuleLoader.Count()) {
+		moduleName = iniModuleLoader.GetModuleName(moduleIndex);
+	}
+	else if (moduleIndex < GetLoadedModuleCount()) {
```
DanRStevens: This check is mildly unintuitive as I was half expecting it to read `consoleModuleLoader.Count()` rather than `GetLoadedModuleCount()`. I suppose I was also expecting some kind of re-indexing (such as `moduleIndex - iniModuleLoader.Count()`), though there is currently no point.

I think I'm going to say no change required. I can't think of any better suggestions here.

srcDLL/op2ext.cpp
```diff
@@ -99,6 +99,12 @@ OP2EXT_API void Log(const char* message)
 	logger.Log(message, FindModuleName(_ReturnAddress()));
 }
 
+
+OP2EXT_API bool IsModuleLoaded(const char* moduleName)
+{
+	return IsConsoleModuleLoaded(moduleName) || IsIniModuleLoaded(moduleName);
```
DanRStevens: For consistency with the method in the later commit, perhaps `IsIniModuleLoaded` should be checked first. It may be good to always checked the two module systems in the same order. In this case I know it really doesn't matter though. It might even make slightly more sense from an efficiency point to check in the order given here, as expected usage is probably with console modules.

I'm going to say no change required.

Brett208:
Thanks for the review. I mostly wanted to check if the external interface looked good as this is much harder to change in the long term. 

-Brett

PR #80, opened 2019-04-25
Add CircleCI config

### LINKED ISSUE #68: Run mingw unit tests on CI
Pulled from issue #47. All research done by @DanRStevens. Previous issue had significant discussions about unit test implementation that wasn't relevant to mingw work.

---

As this is a Windows only project, we'd need to use Mingw to compile from Linux. This requires a bit of special setup for Google Test. In particular, when building Google Test, we must set the command line option to cmake: `-DCMAKE_SYSTEM_NAME="Windows"`. Additionally there are some pthread issues when compiling gtest with Mingw. These are avoided by running cmake with the command line option `-Dgtest_disable_pthreads=ON`. These options can be added to the makefile.

There are additional complications when compiling the test code (as opposed to Google Test itself). To reference the Google Test library, we use `#include <gtest/gtest.h>`. The system installed headers are found in `/usr/include/gtest/`, hence we need an include directory of `-I/usr/include`. Unfortunately, this brings in standard Linux system header files, which conflict with the Mingw specific header files for Windows cross compilation. To get around this, we can change the include directory to be more specific with `-I/usr/include/gtest/`, and then update the source code inclusion to `#include <gtest.h>`. Unfortunately, Google Test itself includes it's other header files by using `gtest/` as part of the path, so those headers won't be found. The solution I found was to copy the `/usr/include/gtest/` folder to another location, perhaps `.build/include/gtest/`, and then add `-I.build/include/` as an additional include folder. That way it brings in gtest, but not the rest of the Linux system libraries.

This would be easier if Google Test didn't depend on the parent folder of the `gtest/` folder being in the include path.

DanRStevens:
This adds a Dockerfile with Mingw and Wine, suitable for doing cross compiles from Linux to Windows, and running the resulting binaries. It then adds CircleCI config which uses this Dockerfile to compile the Windows only op2ext project and run unit tests, all from Linux.

Closes #68.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,15 +1,5 @@
-# For building on CircleCI with Google Test
-
-# Build from the root project folder with:
-#   docker build .circleci/ --tag outpostuniverse/circleci-gcc-clang-googletest
-# Push image to DockerHub with:
-#   docker login
-#   docker push outpostuniverse/circleci-gcc-clang-googletest
-
-# Run locally using the CircleCI command line interface:
-#   circleci build
-# Validate .circleci/config.yml file with:
-#   circleci config validate
+# For building and running unit tests on CircleCI
+# See makefile from project root for local usage
 
 FROM ubuntu:18.04
 
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -24,6 +24,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     mingw-w64 \
   && rm -rf /var/lib/apt/lists/*
 
+# Install wine so resulting unit test binaries can be run
+RUN dpkg --add-architecture i386 && \
+  apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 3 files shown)


DanRStevens:
@Brett208: This is kind of a Linux only change, using largely Linux only tools, though it is cross compiling for Windows, so I thought maybe I'd give you a chance to look through it before merging. If you want.

@cugone: You might find this interesting, given your recent CircleCI work. You're welcome to review it as well. I generally find going over commits individually to be easier than looking at the branch as a whole.

cugone:
This is mostly required Linux Docker boilerplate so I can't really comment on it.

The circleCI file looks fine to me. Implementing executors, caching, and custom jobs isn't needed yet anyway.

DanRStevens:
I suppose Docker kind of obviates the need for caching and custom jobs. All dependencies can be pre-installed to the Docker image. The Docker image is effectively a cache of the build environment. The Dockerfile is the instructions to re-build that cache.

Merging.

PR #87, opened 2019-04-27
Catch filesystem errors

### LINKED ISSUE #78: filesystem throws an exception when non-experimental version used
When compiling with the non-experimental version of filesystem, I am getting the exception below in both Debug and Release builds. 

Directory_Iteratory_Exception

If I force it to build with the experimental filesystem include and namespace using, then it works properly.

Unless an easy fix is available, I would like to regress this change so we can get a release out. I didn't catch this until Monday and don't know what to do to fix.

-Brett

Brett208:
If this looks good, it closes #78. 

Thanks for the help with this one.

-Brett

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -9,7 +9,7 @@
 #include <algorithm>
 
 
-void LocateVolFiles(std::string relativeDirectory = "");
+void LocateVolFiles(const std::string& relativeDirectory = "");
 
 // Declaration for patch to LoadLibrary, where it loads OP2Shell.dll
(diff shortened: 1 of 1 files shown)


Brett208:
@DanRStevens you can see details in issue #78.

The final version of filesystem now throws a filesystem_error (child of std::exception) if a directory_iterator is called on a non-existent directory. Previously, it would just skip out of the directory_iterator as if the directory were empty. 

exception::what is below:

"directory_iterator::directory_iterator: The system cannot find the path specified.\r\n: "C:\ . . . \Outpost2\Addon""

The addon directory is not present by default in Outpost 2, but we still search there for vol files.

This change in behaviour is probably worth investigating in OP2Utility/Archive/MapImager to make sure we are happy.

PR #91, opened 2019-05-01
Add new module tests

Brett208:
I wrote some tests up to test the cross-module code.

This teased out a bug in how the console module's name is returned. Otherwise, things look okay.

Getting this one into the base should allow Crow! to try out a test build of op2ext for his mission code, so would like to be higher priority than other work.


Brett208:
Okay, made the suggested changes. I don't like how the console module's name can be so ambigous, but don't know that I want to go down the rabbit hole of locking down the name by removing all variations of trailing and prefixed slashes and dots, among other practical concerns. at this time.

ReadMe.txt
```diff
@@ -83,6 +77,7 @@ To load a custom module with Outpost 2:
  3. Call 'Outpost2.exe /loadmod directoryName' when executing Outpost 2. For example, Outpost2.exe /loadmod multitek2.
  4. Consider creating a .bat (batch) file that allows for loading the module without opening the command prompt.
 
+To avoid undefined behaviour, console modules should only be stored as a child of Outpost 2's root directory (not further nested or outside of the game's root directory). When calling /loadmod, no trailing or prefixed directory separator should be used. IE do not call `/loadmod testmodule\` or `/loadmod .\testmodule`.
```
DanRStevens: I like the negative examples of what not to do, that's some good detail. I would also pair it with a proper example of what to do. To influence behavior, you often need to model correct behavior.

srcDLL/op2ext.h
```diff
@@ -61,12 +61,12 @@ OP2EXT_API void Log(const char* message);
 
 
 // Performs a case insensitive search of loaded module names, returning true if found.
-// A console module's name is the name of the directory the console module is stored in.
+// A console module's name is the relative directory passed in via the /loadmod command.
```
DanRStevens: When stating "relative directory" it's often helpful to state what it is relative to. It could be the current working directory, or the executable's directory. Often the two are the same, but not always. Ex:
"A console module's name is the folder name parameter passed in via the /loadmod command. The folder name should be relative to the executable's folder, with no trailing slash"

srcDLL/op2ext.h
```diff
@@ -61,12 +61,12 @@ OP2EXT_API void Log(const char* message);
...
 // An ini module name is the module's [section name] within the ini file.
 OP2EXT_API bool IsModuleLoaded(const char* moduleName);
 
 // Performs a case insensitive comparison of the loaded console module name.
-// A console module's name is the name of the relative directory the console module is stored in.
+// A console module's name is the relative directory passed in via the /loadmod command.
```
DanRStevens: Maybe we can reference an earlier comment or help section, rather than repeating the explanation.

srcDLL/op2ext.h
```diff
@@ -80,7 +80,7 @@ OP2EXT_API size_t GetLoadedModuleCount();
 // Retrieves the module name at the specified index. 
 // Ini modules are indexed first and console module is last.
 // Use function GetLoadedModuleCount to determine how many module names to check. 
-// A console module's name is the name of the relative directory the console module is stored in.
+// A console module's name is the relative directory passed in via the /loadmod command.
```
DanRStevens: Ditto.

PR #92, opened 2019-05-03
Comment on Visual Studio Debug Assertion in Test Module

### LINKED ISSUE #88: MSVC reports Debug Assertion when calling free
When using the TestModule, the following debug assertion is thrown by Visual Studio. Only when compiled in debug mode. It occurs even if switched to the delete keyword instead of free. (line 56 below)

DebugAssertion

I'm at a bit of a loss as to why this is happening.

Brett208:
Closes #88


Brett208:
See #94

Issue #100, opened 2019-05-11
Allow `ConsoleModuleLoader` to load multiple modules


DanRStevens:
The `ConsoleModuleLoader` makes use of the global `moduleDirectory` variable, which is referenced by the non-member function `GetArtPath`. This design has several limitations. One being that it complicates testing. Another being this restriction is at the core of why the `ConsoleModuleLoader` only supports a single loaded module.

The `GetArtPath` method is current used to overwrite the two calls from within Outpost2.exe to `CConfig.GetString("DEBUG", "ART_PATH", ...)`. It is effectively hiding the ini file settings:
```
[DEBUG]
ART_PATH = ...
```

Years ago, it was discovered this setting could be used to load resources from the `ART_PATH` setting, outside of the root game folder. The above functionality was added to the console module loader as a way to load mod specific resources in a temporary way, by overriding the ini setting with a module specific folder set on the command line. This is however a bit of a hack, as it's limited to loading resources from a single extra folder. It also prevents the ini file `ART_PATH` system from working, though that setting had been unused before.

----

As we now know more about the resource finding/loading algorithms, perhaps we can design an alternative patch that could allow inserting multiple resource search paths into the game. That would lift the restriction on loading only a single console module, and perhaps allow for easier test code.

One likely place to hook is:
```
00471590  Function: ResManager.GetFilePath(char* resName, [out] char* filePath):bool bSuccess
```
The above method is one of the places where `ART_PATH` is checked. It appears to check the `ART_PATH` location first, then the game folder, then the CD. If we hooked this method, we could insert a search of prioritized module folders before those other locations.

The `GetFilePath` method is called by `ResManager::OpenStream`. If the `GetFilePath` method succeeds, the stream is opened as a file from disk, otherwise it falls back to searching the VOL files. The `ResManager::OpenStream` could be another potential place to hook, however it's called from fewer places, so could not be used to hook as many potential resources. In particular, methods that process level DLLs call `GetFilePath` directly (thought: custom campaigns). As such, the `GetFilePath` method is likely the preferred place to hook.

----

Side note: The other use of `ART_PATH` doesn't seem to be used. There are no known calls or references to the other method, and playing the game does not trigger a breakpoint set on the other method:
```
00471B60  Function: ?(const char* fileName?):bool bSuccess
```
It may be the use of `ART_PATH` in this method is never used, and was hooked simply because it was the only other reference to it.

----

[...]

PR #102, opened 2019-05-11
Add file search path hook

DanRStevens:
This lays the groundwork to address Issue #100.

This is completely untested. I've only verified that it compiles. Nevertheless, I wanted to share this a bit early to get feedback on it (and help testing it).

Note: The branch base is set to something other than master. I did this for ease of review, though we may want to merge the base branch first, and then re-target this PR against master.

### DIFF
--- a/TestModule/DllMain.cpp
+++ b/TestModule/DllMain.cpp
@@ -40,7 +40,6 @@ EXPORT void mod_init()
 	TestPublicInterface();
 
 	TestGetConsoleModDir_s();
-	TestGetConsoleModDir();
 }
 
 EXPORT void mod_run()
--- a/TestModule/TestFunctions.cpp
+++ b/TestModule/TestFunctions.cpp
@@ -44,18 +44,6 @@ void TestGetGameDir()
 	OutputDebugString(gameDirectoryReport.c_str());
 }
 
-// GetCurrentModDir only returns the directory of a module loaded through the console switch /loadmod
-void TestGetConsoleModDir()
-{
-#pragma warning(suppress : 4996)
-	char* moduleDirectory = GetCurrentModDir();
-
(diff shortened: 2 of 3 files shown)


Brett208:
Hopefully my merge commit was okay. Upon testing, code compiles in both debug and release without warnings or errors. 

However, when running, the program has significant problems in both debug and release mode. 

In Release Mode, an exception is thrown after returning from the function in DllMain: 'int __fastcall ExtInit(TApp *thisPtr, int)'.

Right after stepping through this line:
```C++
// Call original function
return thisPtr->Init();
```

This is the runtime exception:
Exception thrown at 0x0047108A in Outpost2.exe: 0xC0000005: Access violation reading location 0x68000FB9.

No warning dialogs are provided by the OP2Memory functions, so it appears to be editing memory without problems.

In Debug mode, Outpost 2 (not op2ext), provides the attached modal dialog with varying instances of no path or garbled paths. It does this for many volume before eventually terminating itself. I didn't check for sure that all volume loads were attempted.

Debug mode ERROR

Let me know how else I can help troubleshoot.

-Brett

Brett208:
Well I really seemed to have gooned this force rebase up. @DanRStevens, probably could use your help fixing. I accidently rebased onto master. Then attempted to rebase onto the consoleModLoader improvement branch, but it doesn't look good.

DanRStevens:
Well, you can use the commit SHA's from the message above:
```
Brett208 force-pushed the addSearchPathHook  branch from 6f6205c to 4b982e4 5 hours ago
```
Checkout the first commit SHA, then create a new branch from it. That should give you a restored branch at the previous state. You can also `reset` the head of the current branch to point to that SHA (check which branch you're on first).

Failing that, there is always [`git reflog`](https://git-scm.com/docs/git-reflog). You should have about a month to restore lost commits before it's actually a real problem.

PR #104, opened 2019-05-16
Search path hook restore

Brett208:
Restored branch for SearchPathHook


DanRStevens:
This shows a lot more history than I was expecting. I'm not too certain what's gone on here, and it's been long enough that I've lost context. I'm going to try doing a rebase from what I have locally and see what the result is.

DanRStevens:
Looks like most of the commits above have already been merged to master. When I ran a rebase on the original branch, it resulted in 2 commits, which were cleanly rebased. Perhaps we should go with that result, in PR #110 while moving forward. Meanwhile the original (PR #102), and this one (PR #104) can be closed.

PR #105, opened 2019-05-16
Improve ini access functions

### LINKED ISSUE #84: Rename `GetOP2PrivateProfileString`
I was looking at `GetOP2PrivateProfileString` and thinking it should be renamed. Perhaps:
`GetOutpost2IniSetting`

That would be more consistent with the existing `GetOutpost2IniPath`.

It also avoids just re-using the Windows method name `GetPrivateProfileString`, which is mostly just an internal implementation detail, and not a very descriptive name.

----

We should wait until open pull requests touching the code settle before doing this.

### LINKED ISSUE #85: Refactor `GetPrivateProfileStdString`
We should refactor `GetPrivateProfileStdString` to remove the internal use of `new` and `delete`. We should instead rely on [`std::string::resize`](https://en.cppreference.com/w/cpp/string/basic_string/resize) to allocate the buffer space.

Additionally, we may consider moving this method out of `FileSystemHelper.h`, which doesn't seem to be a great home for it. I would be fine if the method got its own file.

----

We probably want to write a few unit tests before doing the memory allocation refactor.

Brett208:
Closes #84 
Closes #85

Would like to take a stab at a couple of unit tests before merging if I can add a .ini file to test.exe.

### DIFF
--- a/srcStatic/FileSystemHelper.cpp
+++ b/srcStatic/FileSystemHelper.cpp
@@ -51,7 +51,7 @@ std::string GetPrivateProfileStdString(const std::string& sectionName, const std
 	return profileString;
 }
 
-std::string GetOP2PrivateProfileString(const std::string& sectionName, const std::string& key)
+std::string GetOutpost2IniSetting(const std::string& sectionName, const std::string& key)
 {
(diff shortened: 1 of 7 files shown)


Brett208:
Thanks for reviewing!

-Brett

PR #106, opened 2019-05-20
Allow double spaces in console argument

### LINKED ISSUE #98: Console module directories may fail if containing multiple spaces in a row
This hasn't been tested, but the code within ConsoleArgumentParser appaers to incorrectly concatenate directory strings passed in from the console arguments.

Labeled for fixing before next release is made.

Brett208:
Updated ConsoleArgumentParser so unit tests could act on it without actually passing multiple command line arguments (if that is even possible?). 

Added test cases for ConsoleArgumentParser.

How testing if the ConsoleArgumentParser checks if messages were logged is messy. I couldn't come up with a better solution although I'm sure it exists.

More important, it Closes #98

### DIFF
--- a/srcStatic/ConsoleArgumentParser.cpp
+++ b/srcStatic/ConsoleArgumentParser.cpp
@@ -3,7 +3,6 @@
 #include "FileSystemHelper.h"
 #include "GlobalDefines.h"
 #include <windows.h> // Cannot use WIN32_LEAN_AND_MEAN (it does not contain CommandLineToArgvW)
-#include <system_error>
 #include <vector>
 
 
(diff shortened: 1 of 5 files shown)


Brett208:
If we are testing string equivalency, I'd prefer to expose the error string as a variable from the emitting code base for checking against.

Clearing the log file between tests might be a cleaner approach since you wouldn't have to record a value of some kind before performing each test.

-Brett

DanRStevens:
Hmm, my example was maybe a bit poor. I'd recommend against testing for exact string equivalency with user readable error messages. User readable strings can be updated for wording and clarity, and you don't want corresponding unit tests to be too brittle. That's one reason why I suggested `HasSubstr`, rather than an equivalency check, though my example string kind of missed the mark. You might actually use just a short key phrase, rather than the whole error message, or multiple `HasSubstr` checks, with various keywords.

Brett208:
If ConsoleModuleLoader can emit the string it would send to the logger (or the majority of it or a key phrase) to the unit test, any changes to the message from within ConsoleModuleLoader would be automatically transferred to the unit test. That way you can change all you want and not worry about if the unit test will pick it up. 

Maybe store the log messages as const protected variables within the class. You could create a ConsoleModuleLoaderTest class that exposed them for testing purposes. Maybe more elegant solutions though.

If someone else does the work though, I wouldn't be picky on the particulars I suppose.

PR #108, opened 2019-05-25
Update ReadMe with 2.2.0 changes

Brett208:
Would like to release and push a new version of op2ext to the official website shortly

### DIFF
--- a/ReadMe.txt
+++ b/ReadMe.txt
@@ -120,6 +120,21 @@ Order of vol file precedence is below:
 Change Log
 ------------------------------------------
 
+Version 2.2.0
+
+This version forces quotation marks ("") when passing a directory into the /loadmod switch. Earlier implementations would auto-combine directories with spaces if no quotation marks used, but only allowed passing paths with a single space in a row. So "one space" would work, but "two  spaces" would fail.
+
+ * Allow passing module directories with the /loadmod switch when the directory contains multiple spaces in a row ("  ")
+ * Allow modules to detect which ini and console modules have been loaded via op2ext. (See op2ext.h for new function declarations). May be called from external modules written in C or C++.
+ * Ensure the log file is always created in the game directory. Earlier, the log file was created in the working environment, which could differ from the game directory, such as when attaching a debugger.
+ * If a module fails to load, provide more details to the user
+ * Fix bug in GetGameDir_s that may throw an exception if a buffer length of 0 is passed in as an argument
+ * Split project into a DLL and static sub-projects to facilitate unit tests
+ * Add partial unit test coverage using gtest framework
+ * Update the post build event to use an environment variable when locating Outpost 2's directory for auto-injecting new builds for testing
+ * Allow selecting non-experimental filesystem include when C++17 is available
+ * Add limited mingw project support for Linux compilations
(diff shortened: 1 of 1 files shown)


ReadMe.txt
```diff
@@ -120,6 +120,21 @@ Order of vol file precedence is below:
 Change Log
 ------------------------------------------
 
+Version 2.2.0
+
+This version forces quotation marks ("") when passing a directory into the /loadmod switch. Earlier implementations would auto-combine directories with spaces if no quotation marks used, but only allowed passing paths with a single space in a row. So "one space" would work, but "two  spaces" would fail.
```
DanRStevens: Small point: The quotes are only needed when the path contains a space. This wording makes it sound like all paths require quotes.

DanRStevens:
I was just looking at Git tags, where there is a "Ver2.2.0-beta". I was thinking, maybe we should avoid the "-beta" suffixes, and just stick with point version numbers. In that case, the "beta" release would be 2.2.0, and the next release would be 2.2.1 (or 2.3.0, or 3.0.0). I think it's fine if the changelog had an entry such as "Not publicly released", "Internal release", or similar. Thoughts?

Brett208:
I'm happy including beta in the release tag as it is more descriptive on what the release is for and follows semantic versioning. I suppose a must practice for me. Why does it bother you?

DanRStevens:
You know, maybe it's just that there are two beta tags with different capitalization but the same version number that kind of annoys me. Actually, despite the similarity, they tag different commits, with different dates.

I noticed GitHub marks the "-beta" version as "pre-release", so that's kind of nice.

I suppose I kind of want to avoid having two "2.2.0" release tags. I guess I question the value of having both a "2.2.0-beta" and a "2.2.0".

I also want to avoid having "-beta1", "-beta2", "-beta3", ... for each release. There's something about that that feels like a lack of confidence in the development and release process. The patch number is already part of the version string explicitly for bug fixing.

Brett208:
The double beta tag was probably caused by me messing up a release and then deleting it on GitHub and reposting. Guess it didn't delete the tag when I deleted the bad release. I wanted to make sure the final release matched with the right commit. If I am remembering correctly I forgot to merge a branch in or something on the first attempt, deleted and redid the release.

I'd like to use beta tags if user feedback is requested and we know it isn't going to be a formal release that makes it into Outpost 2. I may be confident in the development process but I'm also humble enough to know that we will sometimes miss things for various reasons. Untested hardware, unintended use cases, non-technical user perspective, etc. The hard part is getting feedback from any non-developers...

DanRStevens:
I think Git might default to pushing all tags. In that case, you may need to delete the tag from both the GitHub repo and your local repo (and everyone else's local repo), otherwise there is a risk the tag will be pushed up to GitHub again. That may have been what happened.

What about tagging such test releases as Beta, but still bumping the version number for subsequent releases? I think I'd be fine with that.

Brett208:
Okay, I'm good with bumping the version even if previous was a beta. I already released this one as 2.2.0, so maybe better to just go with that this time. If you want to try to force change it to 2.2.1 I don't mind.

-Brett

PR #110, opened 2019-05-27
Add search path hook (rebase)

DanRStevens:
This is a rebase of PR #102 onto master. This relates to PR #104 which was a restoration of that work following other rebase attempts. I was not able to restore work to the now closed PR #102, nor reopen the old PR due to GitHub checks.

As reported in PR #102, there seems to be a bug in the code.

----

Edit: This lays the groundwork for #100.


Brett208:
Thank you for fixing much nicer than I did. For easier reference, I pasted the runtime issue below. I really like this fix as it will allow encapsulating moduleDirectory, which would fix the bad hack I had to use to clear moduleDirectory in unit testing. Please let me know how else I can help.

---

Locally, code compiles in both debug and release without warnings or errors.

However, when running, the program has significant problems in both debug and release mode. (Reported problem is different in release than in debug).

**In Release Mode**, an exception is thrown after returning from the function in DllMain: 'int __fastcall ExtInit(TApp *thisPtr, int)'.

Right after stepping through this line:

```C++
// Call original function
return thisPtr->Init();
```

This is the runtime exception:
Exception thrown at 0x0047108A in Outpost2.exe: 0xC0000005: Access violation reading location 0x68000FB9.

No warning dialogs are provided by the OP2Memory functions, so it appears to be editing memory without problems.

**In Debug mode**, Outpost 2 (not op2ext), provides the attached modal dialog with varying instances of no path or garbled paths. It does this for many volume before eventually terminating itself. I didn't check for sure that all volume loads were attempted.

Debug mode ERROR

57663072-6b8bb780-75c0-11e9-8bc8-a6ab28d5ae68

Let me know how else I can help troubleshoot.

DanRStevens:
Can you re-verify this with the current branch? I'm wondering if the problem might have been related to the previous rebasing, or something touched on by another branch. I tried running this in a Windows VM, and am not able to reproduce the problem.

Brett208:
Sorry for the long wait on responding. I retested and the immediate crash I was experiencing is only caused when the debugger is attached. 

When I load Outpost 2 without the debugger attached, Outpost 2 works as advertised.

However, when I attempt to use the /loadmod switch on a valid directory, Outpost 2 appears to immediately shutdown on opening and provides me with no warning. Since I do not have the debugger loaded, I cannot step through to see where it blows up. If useful, I could sprinkle log messages through the code around the SetArtPath function to see where it derails. The crash only happens if a valid directory is chosen for loadmod. If no directory or an invalid directory is provided, op2ext provides a proper warning and works as advertised.

-Brett

DanRStevens:
The new code is only called from `ConsoleModuleLoader::LoadModule()`, after the directory name checks (not empty, specifies an actual directory), so it makes sense a valid console module directory must be specified to trigger the crash.

Only `ConsoleModuleLoader::HookFileSearchPath()` is called immediately. It sets up the hook for the other methods to be called later. How much later I'm uncertain, so it's not clear the problem is necessarily in `HookFileSearchPath`.

I took another look over the code, and I'm not seeing anything obviously wrong. Perhaps a few log messages will help debug this. Would be nice to have an easier way to test this, though I'm not seeing how at present.

Brett208:
Updated my comment. The debugger has nothing to do with the crash. I was calling /loadmod when the debugger was attached, which caused the crash. When I no longer called /loadmod, Outpost 2 works with or without the debugger attached.

Brett208:
Ran this through the debugger some more.

When I set breakpoints in `bool ConsoleModuleLoader::CallOriginalGetFilePath(const char* resourceName, /* [out] */ char* filePath)` and `bool ConsoleModuleLoader::ResManager::GetFilePath(const char* resourceName, /* [out] */ char* filePath) const` the breakpoint is never hit.

The crash happens somewhere in the guts of Outpost 2 after HookFileSearchPath. 

Outpost 2 will just come up with Failed to open volume (null) a couple of times before crashing. 

I assume nothing changed the behavior of VolList so it is still properly setting volume filenames.

We talked about logging some values, but I don't actually know what would be useful to pull out of the code.

DanRStevens:
It might be I got the calling conventions wrong, or messed up the chained call. Or maybe one of the addresses is wrong, though they were copy/pasted and should be free of typos.

I may need to look at the MSVC assembly output for the new methods to see what's going on.

I should also use OllyDbg to check the crash address listed in your earlier message: `0x0047108A`

DanRStevens:
Ahh, I know what's going on. The addresses listed in `callsToGetFilePath` are the start of the `call` instruction. The `call` instruction has a 1 byte opcode, followed by a 4 byte relative address. We want to patch the relative addresses, but the code is patching starting from the opcode. Effectively we have the addresses wrong by 1 byte.

DanRStevens:
I'm having a few doubts about how we disable memory patching for the unit test project. It doesn't seem like such a foolproof and scalable way to do things. Can't think of a better idea for now though.

Hopefully this works a bit better.

srcStatic/OP2Memory.cpp
```diff
@@ -97,8 +104,16 @@ bool Op2MemSetDword(void* destBaseAddr, void* dword)
...
-	const auto postCallInstructionAddress = loadOffset + callOffset + sizeof(void*);
-	return Op2MemSetDword(reinterpret_cast<void*>(callOffset), reinterpret_cast<std::size_t>(newFunctionAddress) - postCallInstructionAddress);
+	if (memoryCommandsDisabled) {
+		return false;
+	}
+
+	// Verify this is being run on a CALL instruction
+	if (*reinterpret_cast<char*>(callOffset) != 0xE8) {
+		throw std::runtime_error("Op2RelinkCall error: No CALL instruction found at given address: " + AddrToHexString(callOffset));
+	}
```
Brett208: I really like the idea of this check.

srcStatic/OP2Memory.cpp
```diff
@@ -3,13 +3,22 @@
...
 #include <sstream>
+#include <iomanip>
 
 
 bool memoryCommandsDisabled;
 std::size_t loadOffset = 0;
 const std::size_t ExpectedOutpost2Addr = 0x00400000;
 
 
+std::string AddrToHexString(std::size_t addr)
```
Brett208: This function is a good idea

Brett208:
@DanRStevens, what do you think about the commit I made? I think it is an improvement, but I'm not sure why you had chosen to throw an exception.

I am pretty sure I'm compiling the code properly. After logging the error instead of throwing, it indicated the problem more clearly (I think).

Thanks,
Brett

DanRStevens:
Ok, thank you for the check. It seems I was a bit sloppy with my last update to `Op2RelinkCall`, and missed two calls to the updated method that needed corresponding adjustments.

You also make a good point about the error reporting. I should not have been throwing exceptions across a module boundary. Any exceptions raised in the DLL should be caught before returning to Outpost2.exe. Logging is a much better option.

Brett208:
It looks like we are not out of the woods yet. 

The call at:

```C++
// Replace call to gTApp.Init with custom routine
if (!Op2RelinkCall(0x004A8878, reinterpret_cast<void*>(ExtInit))) {
	return FALSE;
}
```

is still causing the same error... 

Another Call issue

DanRStevens:
Did you update? I changed that line already.

Brett208:
Hmm, my code base didn't update for some reason. I think I have the changes now, but the result is the same:

try2

I'm not 100% sure that I'm not bungling something up?

Brett208:
Also, I see line 0x004A8877 is:

```
004A8877  |. E8 63ACB50F    CALL OP2EXT.ExtInit                      ;  TApp.Init()
```

Which includes a CALL line (I guess, I'm not really sure what I'm looking at.

DanRStevens:
It looks like it's what it should be. Not sure why an exception is being thrown. Unless maybe that memory was overwritten by another patch?

DanRStevens:
Wait a minute, check that `if (*reinterpret_cast<char*>(callOffset) != 0xE8)` does what we think it does. There might be some integral promotion going on here, with some sign extension. Try changing the `char*` to `unsigned char*`.

DanRStevens:
I came up with what seems like a sensible way to write some unit tests for the `Op2RelinkCall` patching method. It contains a bit of a hack to re-enable and then disable memory patching code. I see that part as possibly temporary in nature. Cleaning up the way we do patching may mean we no longer need such hacks to disable memory writes.

Perhaps other patching methods can have similar unit tests at some point.

Sorry for all the back and forth with untested code. I should have put more effort into figuring out a way to test things myself from Linux. On the plus side, I think some of the casting tricks discovered here may soon allow the project to compile from Linux.

DanRStevens:
I think I know what the problem is. In Issue #100 I noted there were two uses of `ART_PATH`. One is in `ResManager.GetFilePath`, which is hooked in this patch to search additional paths. The other is in `ResManager.CreateStream`, which I had identified in that issue as a mysterious function with no calls or references. It turns out that was an exported method, and a user OllyDbg label had hidden the display of the export info. The call likely comes from OP2Shell.dll, which is why OllyDbg didn't see any calls to it. It seems this other method will also need to be hooked to search additional paths.

I'll work on an update.

DanRStevens:
Hmm, is the behaviour you are checking modified by this pull request? The handling of .VOL files by op2ext is now done independently of Outpost2.exe. The master branch code might not have loaded .VOL files from a module directory. Ever since we reworked the handling of .VOL files to load a custom list, it would no longer have been done through Outpost2.exe methods, and so would no longer be dependent on the `ART_PATH` setting. I think that `ART_PATH` code is from way back before custom .VOL lists, back about when op2ext was first written. If so, perhaps we should file a new Issue. This may be an independent and long standing bug.

As for how to accomplish loading .VOL files from console module folders, we could call `LocateVolFiles(consoleModLoader.GetModuleName())` from within `ExitInit` in DllMain.cpp. I believe that could be added before the "Addon" folder is searched, assuming first added is highest priority.

On another note, I'm looking back at the code and wondering why `LocateVolFiles` only takes a relative path from the executable folder, and can't handle absolute folders. Was this just for easy of concatenation? At any rate, it's not possible to use `consoleModLoader.GetModuleDirectory()` with `LocateVolFiles`.

DanRStevens:
Actually, before going to do that, we may have a problem with directory separators in the replacement `GetFilePath`. In particular, the path to the `moduleDirectory` might not end with a trailing directory separator. That would cause problems for the section:
```
	const auto path = moduleDirectory + resourceName;
	if (INVALID_FILE_ATTRIBUTES != GetFileAttributesA(path.c_str())) {
```

If `moduleDirectory` has no trailing slash, then the string operation will produce an invalid path. The `GetFileAttributes` check will then fail on a non-existent file. The code then continues on assuming there was no such file in the module directory, and falls back to the default search algorithm.

The `moduleDirectory` value is initialized as:
```
	moduleDirectory = fs::path(GetGameDirectory()).append(moduleRelativeDirectory).string();
```

I don't believe there will be a trailing directory separator.

Brett208:
When using ART_PATH to locate resources, things beyond volumes are recognized and found there such as morale.txt, the icon animations, loose maps, etc. I'm afraid we would break a module like multitek 2 by only registering volumes. Granted, it could be worked around easily by repacking all resources into a volume (I think). 

I occasionally check ConsoleModule behaviour to make sure they still allow loading resources from there directory. It is possible I missed checking this on recent commits.

srcStatic/OP2Memory.cpp
```diff
@@ -110,7 +110,7 @@ bool Op2RelinkCall(std::size_t callOffset, void* newFunctionAddress)
 
 	// Verify this is being run on a CALL instruction
 	if (*reinterpret_cast<unsigned char*>(callOffset) != 0xE8) {
-		PostErrorMessage("OP2Memory", __LINE__, "Op2RelinkCall error: No CALL instruction found at given address: " + AddrToHexString(callOffset));
+		PostErrorMessage(__FILE__, __LINE__, "Op2RelinkCall error: No CALL instruction found at given address: " + AddrToHexString(callOffset));
```
Brett208: Thanks for fixing my mistake here using the hard-coded value instead of the __FILE__ macro. I had forgotten about __FILE__.

test/OP2Memory.test.cpp
```diff
@@ -0,0 +1,57 @@
...
+  EXPECT_NE(0x00000000u, callInstruction.relativeAddress);
+}
+
+TEST_F(Op2RelinkCallTest, GoodOpcodeSelfCall) {
+  // Modification should succeed, now calls itself
+  EXPECT_TRUE(Op2RelinkCall(callInstructionAddr, &callInstruction));
+  // Opcode should stay the same
+  EXPECT_EQ(CallOpcode, callInstruction.opcode);
+  // Relative address should point back to itself
+  EXPECT_EQ(-sizeof(CallInstruction), callInstruction.relativeAddress);
```
Brett208: MSVC is providing the following warning: 

Warning	C4146	unary minus operator applied to unsigned type, result still unsigned	test	\op2ext\test\op2memory.test.cpp	46

DanRStevens: Thank you. I had a feeling that might generate a warning. It didn't seem to matter on Mingw, so I left out the cast.

DanRStevens:
Ok, thank you very much for the bug fixing and testing support. And as always, adding the unit tests to the MSVC project.

I'll merge now and then open an issue for the resource loading.

PR #116, opened 2019-08-28
Split makefile into project specific and generic portions

DanRStevens:
I've updated the makefile and split it into two sections. The main makefile contains all the project specific settings. The second part contains all the general rules, which are included by the main makefile. I figure the generic portion could be copied between projects without modification. The generic portion might even be included from a submodule dependency. It contains all the generic rules for out of source builds, with dependency tracking support.

One of the main additions is support for a declarative project syntax. You can declare a project in one line, giving the project a name, it's output file (.exe, .dll, .lib), and a source folder which will be scanned for source files. It automatically creates the needed dependency rules, and adds special targets for building the project, cleaning intermediate and output files, and listing the source files.

Example:
```
$(eval $(call DefineCppProject,op2extLib,op2ext.lib,srcStatic))
```

Configuration support has been added. It's possible to specify a configuration on the command line to select a compiler toolchain, or to set a default with one line in the main makefile. Each configuration will be compiled to separate intermediate folders. This avoids problems of mixing intermediate files from separate toolchains. It also allows for faster rebuilds when switching between toolchains, as it no longer needs to rebuild everything with each switch. Built-in pre-configured toolchains are `gcc`, `clang`, and `mingw`.

Example:
```
make    # Default to mingw toolchain  (output in .build/mingw/)
make config=clang    # Compile with Clang toolchain  (output in .build/clang/)
```

It's now possible to do a local compile of the Google Test library with any of the toolchains, without needing to edit the makefile. This is particularly handy when trying to get a project compiling for a new toolchain. Suppose a project was originally designed for MinGW, and you want to see how much you can get compiling with Clang. Now you can easily rebuild the Google Test library to work with a Clang build, without breaking previous MinGW settings. Unit test projects will automatically pick up a matching compile of Google Test, so versions with identical configurations are linked together.

Example:
```
make gtest    # Compile Google Test library  (output in .build/mingw/gtest)
make gtest config=clang    # Compile Google Test library with Clang  (output in .build/clang/gtest)

make check    # Run tests  (uses mingw build of unit tests and gtest)
make check config=clang    # Run tests (uses clang build of unit tests and gtest)
```


Brett208:
Dan,

This looks good overall. I recommend continuing forward. It looks similar to how I would run MSVC/Visual Studio in the sense that I would specify with a single drop down x86 or x64. I would then use another drop down to specify debug or release. This sounds very similar to just using a switch on the console. Of course you are talking compilers and not just settings.

Some of what you are doing may even be a bit more refined than what I can accomplish in Visual Studio.

Also, I really like that you can switch compiler or other standard use cases without changing files committed to the repository. It should make for a cleaner experience with version control.

I also like the eventual idea of using makefile settings from a submodule. Use of OP2Utility comes to mind where you may want to prevent redefining the same standard settings in both op2mapimager and op2archive. This would sort of mimic what we do with MSVC and what my goal would have been with cmake if we had stuck with it. 

Also correct that I will not be much help with the actual makefile commands, sorry.

I like high level idea of how different projects in Visual Studio link. Although the actual XML and setting choices are frustrating most of the time.

DanRStevens:
Thank you very much for the feedback.

I managed to improve the unit test project support today. I also cleaned up a bit of the dependency stuff.

I think I want to add a target for intermediate files. That way the DLL project can be compiled to .o files, without triggering the link step. The link step is where that project fails, due to name mangling differences between MinGW and MSVC. Having an intermediate target that doesn't fail means the CI builds can compile more of the code, providing better feedback on code quality.

After that, it's just dealing with the Docker and CircleCI commands.


I can't help but wonder if this attempt at declarative syntax may actually turn out to just be a reinvention of CMake. :thinking:

PR #118, opened 2019-09-05
Remove Visual Studio Source and Header Filters

Brett208:
It is easier to work with the headers and cpp files when they are next to each other instead of in different filters. Also, want to sort files by actual directory location in future, not by filters.


Brett208:
There is an option to 'show all files' with Visual Studio. Unfortunately is literally shows all files in the directories instead of just the legitimate program files. I understand wanting to use normal settings for ease of use, but this has been annoying me a long time and I don't seem to be getting used to it. So, I'd like to make an exception on this setting being 'standard'.

PR #120, opened 2019-09-10
Allow internal game modules

Brett208:
First draft allowing Internal Game Modules and unifying ModuleLoader.

Internal modules currently have to be added to a for statement in `ModuleLoader` for registration. Maybe we can improve this in the future so it is more automatic, but for now part of creating a new internal module would be registering it in the `ModuleLoader`.

Documentation needs to be updated if this looks good.

Bringing in console modules remains outstanding, but I thought there was enough benefit to propose the change in its current form.

Use of the .ini file has changed. I propose a key named InternalModules and another key named ExternalModules (Previously named LoadAddons). Internal modules that you want to load would be listed under InternalModules. Same with external modules (formerly ini modules).

I did my best to make the commits easy to follow...


Brett208:
We have the following warning from Linux automated build:

```
In file included from test/ModuleLoader.test.cpp:3:0:
/usr/src/googletest/googletest/include/gtest/gtest.h: In instantiation of 'testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = int; T2 = unsigned int]':
/usr/src/googletest/googletest/include/gtest/gtest.h:1459:23:   required from 'static testing::AssertionResult testing::internal::EqHelper<true>::Compare(const char*, const char*, const T1&, const T2&, typename testing::internal::EnableIf<(! testing::internal::is_pointer<T2>::value)>::type*) [with T1 = int; T2 = unsigned int; typename testing::internal::EnableIf<(! testing::internal::is_pointer<T2>::value)>::type = void]'
test/ModuleLoader.test.cpp:10:2:   required from here
/usr/src/googletest/googletest/include/gtest/gtest.h:1392:11: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (lhs == rhs) {
       ~~~~^~~~~~
```

Honestly, I'm not sure how to read it...

srcStatic/ModuleLoader.cpp
```diff
@@ -1,58 +1,27 @@
...
-	bool result = true;
+	const auto sectionNames = GetModuleNames("ExternalModules");
 
-	for (IniModuleEntry moduleEntry : modules)
+	for (const auto& sectionName : sectionNames)
 	{
-		if (!CallModuleDestruction(moduleEntry)) {
-			result = false;
+		try {
+			RegisterModule(static_cast<std::unique_ptr<GameModule>>(std::make_unique<IniModule>(sectionName)));
```
DanRStevens: Hmm, right, `unique_ptr` doesn't play very well with inheritance. I'm kind of thinking we should go with convention about ownership, rather than enforcing strict `unique_ptr` semantics. It would also allow us to keep using reference syntax instead of pointer syntax.

test/ModuleLoader.test.cpp
```diff
@@ -1,18 +1,47 @@
...
 #include <gtest/gtest.h>
+#include <memory>
 
-TEST(ModuleLoader, InitializeNoModules)
+
+TEST(ModuleLoader, NoModulesLoaded)
+{
+	ModuleLoader moduleManager;
+	EXPECT_EQ(0, moduleManager.Count());
+	EXPECT_EQ("", moduleManager.GetModuleName(0));
```
DanRStevens: Should this be "" or an exception? For the exported C interface, there are no exceptions, so it made sense to return "" there instead.

test/ModuleLoader.test.cpp
```diff
@@ -1,18 +1,47 @@
...
+	EXPECT_TRUE(ipDropDown == nullptr);
 	
-	EXPECT_EQ("", moduleLoader.GetModuleName(0));
-	EXPECT_EQ("", moduleLoader.GetModuleName(1));
+	EXPECT_EQ(1, moduleManager.Count());
 
-	EXPECT_FALSE(moduleLoader.IsModuleLoaded(""));
-	EXPECT_FALSE(moduleLoader.IsModuleLoaded("TEST"));
+	// Check module name search is case insensitive
+	EXPECT_TRUE(moduleManager.IsModuleLoaded("IPDROPDOWN"));
```
DanRStevens: If modules names are case insensitive, we need to be consistent about that treatment. I don't believe internal code always treats them as case insensitive.

srcStatic/ModuleLoader.cpp
```diff
@@ -85,38 +54,56 @@ std::vector<std::string> ModuleLoader::GetModuleNames(const std::string& moduleT
...
+	if (newGameModule == nullptr) {
+		return;
+	}
 
-	if (dllHandle == 0) {
-		throw std::runtime_error("Unable to load DLL " + dllName + " from ini module section " +
-			sectionName + "." + GetLastErrorStdString(TEXT("LoadLibrary")));
+	for (const auto& currentGameModule : modules) {
+		if (newGameModule->Name() == currentGameModule->Name()) {
+			PostErrorMessage(__FILE__, __LINE__, "You may not add a module with an existing name. Duplicate copies of module name " + newGameModule->Name() + " found.");
```
DanRStevens: Here the module names are being compared in a case sensitive way.

srcStatic/ModuleLoader.cpp
```diff
@@ -1,14 +1,30 @@
...
 	RegisterExternalModules();
 }
 
+void ModuleLoader::RegisterInternalModules()
+{
+	const auto internalModuleNames = GetModuleNames("InternalModules");
+
+	for (const auto& moduleName : internalModuleNames) {
+		if ("ipdropdown" == ToLower(moduleName)) {
+			RegisterModule(static_cast<std::unique_ptr<GameModule>>(std::make_unique<IPDropDown>()));
```
DanRStevens: The boilerplate `if` statements needed here will likely grow quite substantially as more modules are added. I suspect there is an easier way to handle this with templates, and probably with a tuple listing all the built in module classes.

srcStatic/ModuleLoader.cpp
```diff
@@ -40,6 +27,21 @@ void ModuleLoader::RegisterExternalModules()
...
+	const auto isModuleRequested = ToLower(GetOutpost2IniSetting("InternalModules", moduleName));
+
+	if (isModuleRequested == "yes") {
+		return true;
+	}
+	else if (isModuleRequested == "no") {
+		return false;
+	}
+
+	PostErrorMessage(__FILE__, __LINE__, "Internal Module named " + moduleName + " is either missing from the .ini file or contains an innapropriate setting. It must be set to Yes or No");
```
DanRStevens: We should probably only log an error for a bad value, not for a non-existent value. If we log for non-existent values, then commenting the line to disable a module would generate an error message.

test/ModuleLoader.test.cpp
```diff
@@ -1,8 +1,20 @@
...
 #include <gtest/gtest.h>
 #include <memory>
+#include <stdexcept>
 
+// Use to test that two modules with the same name but different casing is rejected from registering with the ModuleLoader
+// Windows file system and .ini key value pairs are case insensitive
+class DifferentCasedNameModule : public GameModule
+{
+public:
+	DifferentCasedNameModule() : GameModule("ipdropdown") {}
```
DanRStevens: It may make more sense to pass the name in as a constructor parameter. Then you can create two instances of the test class with the same name but different cases. That way you're not forced to match up with some real module that gets borrowed for this test case and has no real relation to it.

```
Class DummyTestModule : public GameModule
{
public:
    DummyTestModule(const char* moduleName) : GameModule(moduleName) {}

    void Load() override {}
    bool Unload() override { return true; }
};

// ... 
// Names are case insensitive and must be unique
moduleManager.RegisterModule(DummyTestModule("DummyTestModule"));
moduleManager.RegisterModule(DummyTestModule("dummytestmodule")); // Error
```

test/ModuleLoader.test.cpp
```diff
@@ -43,6 +55,10 @@ TEST(ModuleLoader, InternalModulePassed)
 
 	EXPECT_EQ("IPDropDown", moduleManager.GetModuleName(0));
 
+	// Ensure ModuleManager does not allow multiple modules with same name but different casing
+	EXPECT_NO_THROW(moduleManager.RegisterModule(static_cast<std::unique_ptr<GameModule>>(std::make_unique<DifferentCasedNameModule>())));
```
DanRStevens: I believe this should be `EXPECT_THROW`, with the exception type listed.

srcStatic/IniModuleLoader.cpp
```diff
@@ -8,11 +8,11 @@
 // Load all active modules specified in the .ini file
 void IniModuleLoader::LoadModules()
 {
-	std::vector<std::string> sectionNames = GetSectionNames();
+	const auto moduleNames = GetModuleNames("ExternalModules");
```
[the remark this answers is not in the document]
DanRStevens: Load order can simply be controlled by the order that we check for INI file keys and try to load the modules. Effectively we'll have an internal hardcoded list for the order.

For dependencies, we can force turn off certain modules if a dependent module has not been enabled.

Load order and dependencies may become more relevant later as we clean up some of the patch code. I suspect some of the patches can be split to be more targeted or more reusable, which may reveal a few order dependencies that have been kind of hidden so far with how a lot of the patches are all kind of mushed together into op2ext with no clear separation.

This is more of a concern for future development than anything actionable right now.

----

More immediately actionable is how we might name INI keys. I'd like fairly standard naming convention if possible. Not all patches do quite the same thing, so we may need some slight but controlled variation in the names. For example:
```
; Disabling code
DisableCdCheck = yes
DisableCheats =  yes
; Enabling new features
EnableIpDropdown = yes
EnableUnitLimitIncrease = yes
EnableGarageSaveLoadFix = yes
```

srcDLL/DllMain.cpp
```diff
@@ -64,7 +64,7 @@ int __fastcall ExtInit(TApp *thisPtr, int)
 		PostErrorMessage(__FILE__, __LINE__, stringStream.str());
 	}
 
-	InstallIpDropDown();
+	ipDropDown.Initialize();
```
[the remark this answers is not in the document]
DanRStevens: Sorry, I wasn't very clear here. We seem to have a reversed concept of dependencies between static and member method.

One of the possibilities I was thinking of is:
```
auto gameModule = IpDropDown::Patch(); // Call static method which creates the object
```
As opposed to creating the object first, and then patching:
```
IpDropDown ipDropdown;
ipDropDown.Initialize();
```

Another possibility, different from the above, is to use an RAII technique:
```
if (IsInternalModuleRequested("IpDropDown")) {
    RegisterModule(new IpDropDown());  // Constructing the object activates the patch
}
```
In this case, the patching code does go in the constructor, and we have no need for the `Load()` method. (Contrary to the documentation we recently wrote about not patching from the constructor, because objects in the current design are constructed before they are enabled / registered).

My idea of using tuples may be a bit premature, and perhaps not well thought out. You can define them to hold arbitrary types, though they are still constructed at runtime with all types instantiated. Perhaps what I want is closer to what I saw in the Google Test documentation. I don't actually know how they do it, but clearly they were able to do it. It was the section on [Type Parameterized Tests](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#type-parameterized-tests):

> Finally, you are free to instantiate the pattern with the types you want. If you put the above code in a header file, you can #include it in multiple C++ source files and instantiate it multiple times.
> ```
> typedef ::testing::Types<char, int, unsigned int> MyTypes;
> INSTANTIATE_TYPED_TEST_SUITE_P(My, FooTest, MyTypes);
> ```
> 

Perhaps it may require evil macros to accomplish.

srcStatic/ModuleLoader.cpp
DanRStevens: As in, we assume the module loader has ownership of all registered modules, and it's free to delete / destruct those objects. The compiler won't try to clear out the pointers of a caller when transferring ownership, though if the caller happened to hang on to a pointer, it may risk a crash if it called methods after the `ModuleLoader` had destructed that module. That's not to say a stored module pointer could never be used, just we'd need to have a clear definition of when the `ModuleLoader` would destroy the object to determine when it may be safe to call methods on the module.

```
auto module = new SomeGameModule();
module.Method(); // Sure, this is fine.

moduleLoader.Register(module); // Transfer ownership to module loader (by convention)

module.Method(); // This might be unsafe, if moduleLoader has destroyed the object

moduleLoader.UnloadModules(); // Though the module loader might specify it's safe until now

module.Method(); // Undefined behavior. Likely a crash. Method called on destroyed object

delete module; // Bad! You don't have ownership. This is moduleLoader's responsibility.
```

More likely though, the code could end up looking something like:
```
moduleLoader.RegisterModule(new SomeGameModule());
// No local pointer available to call methods, or potentially do anything bad with
```

test/ModuleLoader.test.cpp
DanRStevens: Yeah, I think this should probably be an exception. It's trying to access data about a non-existent module. There should be no reason for that last line to ever be executed. Especially since the first line allows code to guard against it, and more likely, loop mechanics would prevent it from happening, because really, a method like that is likely only going to be called from some kind of loop.

srcStatic/GameModules/GameModule.h
```diff
@@ -11,10 +11,11 @@ class GameModule
 	inline std::string Name() const { return name; };
 
 	virtual void Load() = 0; //mod_init(), InitMod
-	// Only available for Console modules (other modules would just call an empty function)
-	//virtual void Run() = 0; //mod_run()
 	virtual bool Unload() = 0; //mod_destroy(), DestroyMod
 
+	// Only available for Console modules (other modules just call an empty function)
+	virtual void Run() { }; //mod_run()
```
DanRStevens: This documentation is probably more suited to the derived class. It's the derived class that potentially overrides this method and decides if it actually does anything. Adding support for this in INI modules would not require changes to the base class. It would only need to define a DLL method to call, and then start calling it from the derived class that handles INI modules.

test/ModuleLoader.test.cpp
Brett208: I don't think we should create an exception for adding 2 modules with the same name. We would have to catch the exception to keep Outpost 2 from closing. I think notifying the user of the issue and logging is sufficient.

srcStatic/ModuleLoader.cpp
Brett208: I was thinking about this after committing and came to the same conclusion. I'll change in the near future.

test/ModuleLoader.test.cpp
DanRStevens: Ahh, ok, I understand. I would be tempted to catch the exception and translate it to a log message, but it depends on how deep down the code is. I don't really like having to catch exceptions directly around where they are generated.

test/ModuleLoader.test.cpp
```diff
@@ -55,11 +55,19 @@ TEST(ModuleLoader, InternalModulePassed)
...
-	// Ensure ModuleManager does not allow multiple modules with same name but different casing
-	EXPECT_NO_THROW(moduleManager.RegisterModule(static_cast<std::unique_ptr<GameModule>>(std::make_unique<DifferentCasedNameModule>())));
-	EXPECT_EQ(1u, moduleManager.Count());
-
 	EXPECT_NO_THROW(moduleManager.LoadModules());
 	EXPECT_NO_THROW(moduleManager.RunModules());
 	EXPECT_NO_THROW(moduleManager.UnloadModules());
 }
+
+TEST(ModuleLoader, RejectCaseSensitiveNames)
```
DanRStevens: The test name is not entirely clear and maybe a little misleading. A string by itself is not really case sensitive nor case insensitive. It's a comparison of two or more strings that can be done with regards to case sensitivity. Related, though not relevant here, is case folder versus case preserving. That would apply to how a single string is handled, and may affect how a comparison might be done. Here, registering a module name is case preserving, while comparisons are case insensitive. The phrasing for what we want to express feels a bit awkward. Maybe "RejectCaseInsensitiveDuplicateNames"?

The test itself is wonderfully clear. I like how you separated the checks into their own test case. The code is very targeted and easy to read.

test/ModuleLoader.test.cpp
Brett208: OK, done

srcStatic/ModuleLoader.cpp
Brett208: If we want to allow access to the module outside of ModuleLoader after it has been registered, we could use a shared_ptr. I wasn't sure we wanted to go with a unique_ptr, but it seemed to be the easier to test and maintain solution so I went with it. If we want modules to be manipulated after registration, I wouldn't be opposed, but was sort of waiting to see if a use case appeared before supporting it.

I would prefer a shared or unique pointer to a raw pointer because they tell the user more about the intention and working of the pointer without having to review comments or conventions. It also reduces the chance of trying to de-allocate memory incorrectly by allowing the standard library to cover it.

I suppose if it was proven that we could not accomplish our goals for how the ModuleLoader works using a shared or unique pointer, I would be amenable to using a raw pointer, but I haven't seen that yet.

srcDLL/DllMain.cpp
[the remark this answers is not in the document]
Brett208: OK, the constructor portion makes sense now. I agree there is nothing stopping us from executing the Load function from the constructor and the unload in the destructor. This would probably involve removing the virtual Load and Unload function from GameModule and instead executing its code in the constructor and destructor. There is a certain symmetry to this solution.

I'm happy with how things are, but am okay with reprogramming to to use constructor/destructor instead of virtual load and unload function. @DanRStevens would you like to see it changed constructor/destructor? Shouldn't be too hard...

---

OK, I see what you mean that we could use a static function to create the object and return it, but I don't feel like it adds much here?

---

For the tuple information, I think what you are describing looks generally good to me, but it is still fuzzy. I sort of understand the idea of tuples based on some C# work I've done, but suspect there are differences in how C++ implements. I like the idea of punting for now on this for now in an issue or just waiting to see if it becomes useful later.

Brett208:
I'm done with this round. It looks like there is at least one test case name for me to look at next round.

We need to come to consensus on what to call the section names in the .ini file. I can accept [GameAddons] instead of updating to [ExternalModules] if backwards compatibility is considered a big enough deal. Since most people will update op2ext by downloading the newer version of Outpost 2 when we post it, I think the fallout would be minimal, especially if we warn of it in the changelog / release documentation. I do feel like [ExternalModules] is more descriptive, especially in light of adding the concept of internal modules. 

For internal modules, I would like to avoid the term [Patches] by itself. An INI or Console module could be a patch as well so it isn't clear what belongs in either section. If the term patch is preferred to modules, perhaps [InternalPatches]? 

Thanks for the feedback so far.

test/ModuleLoader.test.cpp
Brett208: Thank you for reviewing.

srcStatic/ModuleLoader.cpp
DanRStevens: Currently the use of `unique_ptr` requires some pretty ugly casts at the point of call. That's what I really want to avoid. I don't think we really need to switch to `shared_ptr`, as that would leave the problem the same.

I'm not opposed to using `unique_ptr` internally to manage lifetime. I'm mostly just concerned about the impact it has on the API. I would like an API where the caller doesn't have to cast to call a method.

Supporting manipulation of the modules after loading doesn't seem particularly important. It's more that not using a `unique_ptr` makes that a possibility that should be considered.

----

Bit of a side note here.

There is also another API cost to using `unique_ptr`, which ties in with taking ownership of objects (whether or not they're using `unique_ptr`). When you take ownership in a way that guarantees `delete` will be called on the object pointer at some point, you can no longer pass global or local variables. It would be an error to try to `delete` a global or a local variable.

The way some of the code is currently written, a few of our patches are written as global objects. That is however the ad hoc nature of existing patch code. As we rewrite things, that will change. Hence, this is just a side note, not really a show stopper.

With the current plan going forward, particularly if we use RAII and apply patches in the constructors and remove them in destructors, it does make sense to create the patches as dynamic objects and fully manage their lifetime.

srcDLL/DllMain.cpp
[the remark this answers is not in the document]
DanRStevens: Yes, this might be more of a longer term plan. It would require messing around with templates, which can sometimes be tricky to get right.

Moving from `Load`/`Unload` to constructor/destructor patching shouldn't be too difficult of a change, from the module perspective, so it's fine if we punt on that too.

Actually, to really make use of the constructor/destructor patching, we could design some RAII type patch objects. Such objects could patch/unpatch common instructions in their constructor/destructor. That would allow a very declarative syntax to patch calls and jumps, which are the majority of patch code. Example:
```
DisableCdCheckPatch : GameModule {
private:
  JumpNop<0xsomeaddr> neverJumpToCdCheckCode;
  JumpUnconditional<0xsomeaddr> alwaysSkipCdCheckCode;
};

GarageSaveLoadPatch : GameModule {
public:
  // Constructor
  GarageSaveLoadPatch() :
    // Initialize RAII patch objects
    preSaveTranslate(&GarageSaveLoadPatch::NewPreSaveTranslate);
  {}

private:
  bool NewPreSaveTranslate(params) {
    // Patch code ...
  };

  CallRelink<0xsomeaddr> preSaveTranslate;
};
```

srcDLL/op2ext.cpp
```diff
@@ -126,13 +126,19 @@ OP2EXT_API size_t GetLoadedModuleName(size_t moduleIndex, char* buffer, size_t b
...
+			moduleName = "";
+		}
 	}
-	else if (moduleIndex < GetLoadedModuleCount()) {
-		moduleName = consoleModLoader.GetModuleName();
-	}
-	else {
+	catch (...) // Prevent throwing an error across DLL boundaries
+	{
 		moduleName = "";
```
DanRStevens: The variable `moduleName` is effectively initialized to `""` 3 times in this method. We can remove the later two:
```
std::string moduleName; // Implicitly initialized to ""
moduleName = ""; // Else block (redundant)
moduleName = ""; // Catch block (redundant)
```

The code in the try block has strong exception safety (commit or rollback semantics). That is, if an exception is raised, the code is guaranteed to have no side effects. The only possible side effect is assigning to `moduleName`, but this is done after all possible exceptions are raised. Hence if an exception is raised, the value of `moduleName` should still be it's initial value of `""`.

Similarly, if the `else` block executes, then we've skipped over the other assignments of `moduleName`.

----

We should probably add some logging in the `catch` block. It would be better if errors like this were not silently ignored.

Brett208: I figured you wouldn't like the changes to this function after writing it without logging.

Sometimes I think of string being an object in C# terms which means it isn't auto-initialized. Good point on the redundant else statement. I'll pull it out and add some sort of logging.

DanRStevens: To be fair, I did like the change, I just noticed we could improve it even further. =)

Interesting to know you already knew what my suggestion would be. From my end it feels like the things I'm able to comment on are forever decreasing in impact.

Brett208: The real question you should be asking is why I didn't just fix it before pushing if I knew it would bother you... :/

Ok, it is coming in the next push.

Brett208:
Agreed we should standardize on the word modules or patches. After thinking about it some more, I'm ok with either the word modules or patches. Technically you could use the console module loader to bring in a module that just swaps out a volume or tech tree or something. But calling this a module vs a patch seems to be splitting hairs to me. Both terms work. 

If we go with patches, we would need to update the code from GameModule.h to GamePatch.h. Probably some readme changes as well, but all easy to do.

I don't prefer having all the modules (patches) in one section of the ini together. Since one requires external files, it is nice to see them separated.

I can live with BuiltInModules or BuiltInPatches. I think it would be clear enough to end users. Also agree with the symmetry comment though. If we go with "BuiltIn", I would want to change a couple of comments and function names from Internal(blah) to BuiltIn(blah).

Maybe we just need to pick a convention and move forward at this point. If you want to talk it out more, we can meet up in IRC? I think we are close enough in agreement if you just want to post your final preference at this point we can go with it (as long as it isn't putting them all in one section together).

---

I missed commenting on it earlier, but it was brought up to log every module that was loaded. I'd like to avoid this for now. It would tend to bloat the log file, esp if we actually get a lot of BuiltInPatches. I wouldn't want to add a dozen lines to the logger saying everything went as advertised every time the program loads. (unless we go with adding a verbose option to the logger specifically for this kind of activity). I would support improved logging of failed to load modules if we think it isn't getting logged sufficiently when things do not work properly.

Thanks,
Brett

srcDLL/op2ext.cpp
DanRStevens: Because not everything needs to be perfect on the first try. Forward progress is still forward progress!

"Forward! Not Backward! ... and always twirling, twirling ..."
https://www.youtube.com/watch?v=0Udvtf77hb0
:wink:

DanRStevens:
I think I would prefer "Modules" over "Patches". The term sounds slightly more inclusive and better encompasses all that can be done.

I think I would prefer "BuiltIn" over "Internal". Would be nice if we had a corresponding word for "External" that maintains symmetry, though perhaps it's actually fine.

I believe the Console modules were designed specifically with the intent of being able to replace resources without having a DLL present. That aspect has so far been maintained, minus Issue #121 relating to replacement VOL files.

I'm fine with splitting the two module types into two INI file sections. It's only really the one setting for Ini Modules right now, but maybe that will change in the future, so it's own section is probably a good idea.

----

You may have a good point about the logging. I guess I considered the log file to be kind of out of the way, and you wouldn't normally look into it unless there was a problem. And in that case you probably do want verbose logging. Actually, the idea of verbose logging is a bit funny in that you have to have already needed it before you turn it on. I suppose that's good enough though if the problem is easily reproducible.

srcDLL/op2ext.cpp
```diff
@@ -134,8 +135,10 @@ OP2EXT_API size_t GetLoadedModuleName(size_t moduleIndex, char* buffer, size_t b
 			moduleName = consoleModLoader.GetModuleName();
 		}
 	}
-	catch (...) // Prevent throwing an error across DLL boundaries
+	catch (const std::exception& e) // Prevent throwing an error across DLL boundaries
 	{
+		logger.Log("op2ext threw an exception attempting to locate and pass the module name for module loaded at index " + 
+			std::to_string(moduleIndex) + ". Details: " + std::string(e.what()));
 		moduleName = "";
```
DanRStevens: This line is still redundant.

Brett208: I am out of time today. I'll take another stab at this hopefully tomorrow. Also, will try to fix the merge commit tomorrow. 

-Brett

srcStatic/ModuleLoader.cpp
DanRStevens: I found this article, which may be relevant here:
[Move smart pointers in and out functions in modern C++](https://www.internalpointers.com/post/move-smart-pointers-and-out-functions-modern-c)

It includes some guidelines for when to use smart pointers in interfaces, including a link to a [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics) document, which examples of when and why not to.

I've been somewhat aware there were downsides, and have been hesitant to use them as part of my own interfaces, though I never really read in detail why, nor assessed if this here is an appropriate use case.

DanRStevens:
Ahh, I see what caused the merge commit. Because IniModuleLoader.cpp was moved, and one line of the file in the old location was changed:
`GetLastErrorStdString(TEXT("LoadLibrary"))` => `GetLastErrorString("LoadLibrary")`

DanRStevens:
I'll take a look at these new merge conflicts.

Brett208:
Ok, I'll work on cleaning up these merge conflicts now.

DanRStevens:
Looks like we both replied at the same time. I'll hold off a bit.

DanRStevens:
Ahh, it was the `return "";` => `return std::string();` change in IniModuleLoader.cpp, and the addition of "GetCommandLineArguments.cpp" to the static library project file.

Brett208:
OK, wording updated. We now have Built-in modules (BuiltInModules) and External Modules.

I'd like to merge now and push the following issues:

1. Switch loading External INI Modules from a single key/value to an ini section (need to develop a way to register all members of a section without knowing their names ahead of time.)
2. Update documentation in readme and sample modules once we finish work on the console modules.

@DanRStevens, if you would like to play with the code some based on your unhappiness with the current use of unique_ptr and std::move in registering modules, please jump in and make your changes. Otherwise, I would be happy if they were pushed to an issue for future work. 

This branch's changes are far reaching and getting in the way of other branches regularly now.

srcStatic/ModuleLoader.cpp
DanRStevens: This does actually seem like an appropriate use of `unique_ptr` as a function parameter.

The real issue with the casting came from pass by reference, rather than pass by value.

Brett208:
Hehe, it would help if I knew how to properly use move semantics. Thanks for fixing. Just tested with IPDropDown module on MSVC debug compilation and it worked.

Issue #113, opened 2019-06-07
Remove dependence on Outpost2DLL


DanRStevens:
Currently op2ext relies on Outpost2DLL to link to two `TApp` methods. This dependence prevents buildings and testing op2ext.dll on Linux due to name mangling differences with the Mingw compiler.

In Issue #100 and PR #102 / PR #110, an alternate linking method was developed to hook the resource search path. (Currently this method needs debugging due to a crash problem). This same method could be applied to the `TApp` linking. That would solve the above Linux compilation issue.

Although the proposed linking code seems to contain a bug, applying the same method to the `TApp` linking code could allow for compiling and debugging of the proposed code on Linux.

Related to the above, the makefile could be updated to support building the DLL or the static lib as separate projects.

DanRStevens:
PR #116 cleaned up the Makefile and added rules to build either the DLL or the static library as separate targets. This removes one of the impediments to working on this issue.

Brett208:
Can you actively run Outpost 2 with op2ext compiled on Linux now?

Thanks,
Brett

DanRStevens:
Yes. Though it does crash upon exiting due to the Mingw code generation problem when it tries to call the original `TApp.ShutDown()`. Annoying, but good enough to run some quick and dirty manual tests. Maybe not yet quite so good for automated tests.

PR #122, opened 2019-09-12
Enable Linux DLL build

DanRStevens:
Adds support for full DLL build from Linux with Mingw.

Edit: This supports Issue #113.

### DIFF
--- a/makefile
+++ b/makefile
@@ -18,7 +18,7 @@ include makefile-generic.mk
 config := mingw
 
 
-CXXFLAGS := -std=c++17 -g -Wall -Wno-unknown-pragmas
+CXXFLAGS := -std=c++17 -O2 -g -Wall -Wno-unknown-pragmas
 LDFLAGS := -static-libgcc -static-libstdc++ -LOutpost2DLL/Lib/
 LDLIBS := -lOutpost2DLL -lstdc++fs -lws2_32
 
--- a/makefile
+++ b/makefile
@@ -39,11 +39,14 @@ $(eval $(call DefineCircleCi))
 
 ifdef Outpost2Path
 
+.PHONY: install run
+
+install: $(Outpost2Path)op2ext.dll
+
 $(Outpost2Path)op2ext.dll: op2ext.dll
 	cp op2ext.dll "$(Outpost2Path)"
 
-.PHONY: run
-run: $(Outpost2Path)op2ext.dll
+run: install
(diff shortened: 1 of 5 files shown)


DanRStevens:
I've been doing a bit of testing on this and have observed crashes on Linux. Oddly, I managed to solve one crash by enabling optimizations. Now it's crashing at a later place in NetHelper, which seems to be trying to dereference a `nullptr`. I'm investigating the cause.

DanRStevens:
Ok, that's really good to hear. This might be a Linux only bug caused by a Mingw incompatibility. It looks like there is some memory corruption happening on Linux. I'll continue to investigate.

DanRStevens:
It seems the faulty handling of `filesystem::is_directory` under Mingw was causing problems with .vol file loading. I applied the fix we discovered elsewhere, and was able to get things to run.

I still experienced crashes during shutdown, but that was traced to bad code generation linking to the `TApp.ShutDown()` call. It seems it was setting the hidden `this` pointer to an uninitialized value from the stack within the local variable section of the stack frame, rather than reading it as the parameter pushed in the `ECX` register. This lead to a crash in the original code when it tried to access member variables. The problem was very clearly caused by Mingw specific code generation.

I replace the `__thiscall` hack with member function casting. Both are MSVC specific hacks, but I think the later is a bit cleaner. This change affected the section of badly generated code, though unfortunately it's still bad. The chained call now loads the correct ECX, but then offsets it by an uninitialized value. The offset is normally to account for things like multiple inheritance, which isn't used here.

I'm stopping here for now.

Assuming nothing broke on Windows, I may want to merge the current changes. It's a big step up on Linux now that the code compiles with Mingw and runs. It's only when it tries to clean up during shutdown that there is still a problem. That's good enough for many development needs.

Brett208:
I compiled and ran these changes in both debug and release modes. Neither seems to produce any warnings or errors on MSVC. I verified that NetFix and IPDropDown appear to load properly. Everything seems to shut down properly. All tests also passed locally for debug mode.

@DanRStevens, unless you have some specific tests you would like me to perform, I'm happy if you merge.

I read through the commits although most of the material felt beyond my application. Too bad mingw requires a hack when using filesystem. I am glad you will have an easier time contributing with it fully compiling on mingw.

DanRStevens:
Thank you very much for the Windows development support.

PR #124, opened 2019-09-18
Remove extra library includes from VS projects

DanRStevens:
I don't think these libraries were ever used by these projects. They may have been there from copying other project files as a template. Some might also have been related to a minimal CRT (C-RunTime) library that was used to replace the traditional one, and so needed explicit references to core system libraries. The minimal CRT library was once used for creating much smaller DLL sizes for mission DLLs.

### DIFF
--- a/TestModule/TestModule.vcxproj
+++ b/TestModule/TestModule.vcxproj
@@ -58,7 +58,7 @@
     </ClCompile>
     <Link>
       <RandomizedBaseAddress>false</RandomizedBaseAddress>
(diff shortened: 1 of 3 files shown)


DanRStevens:
Yeah, it's funny how static libraries work. You don't actually need to import libraries from the static library project. You do however need to import the libraries used by a static library when building an exe or dll with it. I suppose that's so you can delay the final linking until the end, which leaves you more free to swap versions around, assuming the methods all have compatible linkage between the versions.

DanRStevens:
It occurs to me in hindsight that you'd already cleaned up some of the library includes in the other open branch. I guess that's why the changes seemed familiar. There are a few small merge conflicts now. :frowning_face:

PR #126, opened 2019-09-19
Update command line parsing

DanRStevens:
This updates the command line argument processing code to make it more testable. Errors are reported with exceptions, which are now only caught and logged at a higher level. This means unit tests no longer have to muck about with the log file to check for error reporting.

This cleanup is meant to help with Issue #100.

### DIFF
--- a/srcStatic/ConsoleArgumentParser.cpp
+++ b/srcStatic/ConsoleArgumentParser.cpp
@@ -3,6 +3,7 @@
 #include "StringConversion.h"
 #include "FileSystemHelper.h"
 #include "GlobalDefines.h"
+#include <stdexcept>
 
 
 std::vector<std::string> GetCommandLineArguments();
@@ -24,18 +25,22 @@ std::string FindModuleDirectory()
 
 std::string FindModuleDirectory(std::vector<std::string> arguments)
 {
-	const std::string switchName = GetSwitch(arguments);
+	try {
+		const std::string switchName = GetSwitch(arguments);
 
-	if (switchName.empty()) {
-		return std::string();
+		if (switchName.empty()) {
+			return std::string();
+		}
+
+		if (switchName == "loadmod") {
+			return ParseLoadModCommand(arguments);
+		}
+
+		throw std::runtime_error("Provided switch is not supported: " + switchName);
+	} catch(const std::exception& e) {
+		PostErrorMessage(__FILE__, __LINE__, "Error parsing command line arguments: " + std::string(e.what()));
+		return "";
 	}
-
-	if (switchName == "loadmod") {
-		return ParseLoadModCommand(arguments);
-	}
-
-	PostErrorMessage(__FILE__, __LINE__, "Provided switch is not supported: " + switchName);
-	return std::string();
 }
 
 std::string GetSwitch(std::vector<std::string>& arguments)
--- a/srcStatic/ConsoleArgumentParser.cpp
+++ b/srcStatic/ConsoleArgumentParser.cpp
@@ -13,8 +13,13 @@ std::string ParseLoadModCommand(std::vector<std::string> arguments);
 
 std::string FindModuleDirectory()
 {
-	const auto arguments = GetCommandLineArguments();
-	return FindModuleDirectory(arguments);
+	try {
+		const auto arguments = GetCommandLineArguments();
+		return FindModuleDirectory(arguments);
+	} catch(const std::exception& e) {
+		PostErrorMessage(__FILE__, __LINE__, "Error parsing command line arguments: " + std::string(e.what()));
+		return "";
+	}
 }
 
 std::string FindModuleDirectory(std::vector<std::string> arguments)
--- a/srcStatic/ConsoleArgumentParser.cpp
+++ b/srcStatic/ConsoleArgumentParser.cpp
@@ -25,22 +25,17 @@ std::string FindModuleDirectory()
 
 std::string FindModuleDirectory(std::vector<std::string> arguments)
 {
-	try {
-		const std::string switchName = GetSwitch(arguments);
+	const std::string switchName = GetSwitch(arguments);
 
-		if (switchName.empty()) {
-			return "";
-		}
-
-		if (switchName == "loadmod") {
-			return ParseLoadModCommand(arguments);
-		}
-
-		throw std::runtime_error("Provided switch is not supported: " + switchName);
-	} catch(const std::exception& e) {
-		PostErrorMessage(__FILE__, __LINE__, "Error parsing command line arguments: " + std::string(e.what()));
+	if (switchName.empty()) {
 		return "";
 	}
+
+	if (switchName == "loadmod") {
+		return ParseLoadModCommand(arguments);
+	}
+
+	throw std::runtime_error("Provided switch is not supported: " + switchName);
 }
 
 std::string GetSwitch(std::vector<std::string>& arguments)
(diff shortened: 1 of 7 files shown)


DanRStevens:
I added a few changes to the new `LocalResource` template class I introduced in this PR. The changes here will make it easier to re-use the new class to further cleanup other code. Namely, the `GetLastErrorString` method which was refactored in PR #127, and which also needs to use `LocalFree`.

srcStatic/ConsoleArgumentParser.cpp
```diff
@@ -29,7 +29,7 @@ std::string FindModuleDirectory(std::vector<std::string> arguments)
 		const std::string switchName = GetSwitch(arguments);
 
 		if (switchName.empty()) {
-			return std::string();
+			return "";
```
Brett208: I specifically wrote std::string() to indicate an empty string and prefer it. Although I don't mind if you change it. I have been inconsistent with the application in other parts of the code base. With the quotes, I feel like I don't know for sure you are returning a std::string without looking at the function return type.

https://stackoverflow.com/questions/26587110/returning-an-empty-string-efficient-way-in-c/26588207

test/ConsoleArgumentParser.test.cpp
```diff
@@ -38,20 +36,3 @@ TEST(ConsoleArgumentParser, NoSwitchArgument) {
 TEST(ConsoleArgumentParser, TooManyArguments) {
 	EXPECT_THROW(FindModuleDirectory(std::vector<std::string> { "/loadmod", "path1", "path2" }), std::runtime_error);
 }
-
-
-// Returns true if calling FindModuleDirectory provides input to the logger by checking log size
-::testing::AssertionResult FindModuleDirectoryIsLogged(
```
Brett208: I wanted to see this morph into looking for strings on the last logged line at some point to better test PostErrorMessages. I suppose it can be re-added in the future if we implement it in that form.

srcStatic/ConsoleArgumentParser.cpp
DanRStevens: Hmm, very interesting link. It seems `std::string()` and `{}` compile to the same, while `""` can compile to less efficient code under certain compiler optimization settings. I checked it out for a few compilers and settings on: https://godbolt.org/

I think I'll switch it back to `std::string()`. That's certainly more readable than `{}`.

test/ConsoleArgumentParser.test.cpp
DanRStevens: I was thinking about that sort of thing too. This might be a good problem to try Google Mock on. If we can create a mock object for the logger, than we can verify more directly if we are logging proper error messages, without needing to actually go to disk.

To test with mock objects (with or without Google Mock), we'd need to make a few changes to the logging code. We'd have to split the `Logger` class into an interface class, and a separate implementation. This might be a good idea to do anyway, since it would allow for multiple logging output options.

We would also need to change the global `logger` variable to be a pointer or reference to a `Logger`. That would allow method calls to go through to a potential derived class. I'd recommend a pointer if we want to allow for updates to the logging output at runtime, allowing it to be swapped out for a different type of logger. A global reference can only be set once, and it would be done during global initialization.

Not required, but something I was thinking about. Perhaps we should use global logging functions, rather than a global logging object. The global functions can then delegate to the logging object internally. That would allow restricting access to the logging object, and possibly making it not a global. That could have benefits for deferring the creation of the logging object. It would also protect the code from having to change `logger.` to `logger->` all over the place if we change it from a `Logger` instance to a pointer.

Finally, with Google Mock / Google Test support, we might potentially use the string matchers on the error messages. That would allow for testing of keywords, rather than exact error messages. Small changes in the error messages wouldn't immediately break tests, and we also wouldn't have to export the error strings to give tests access. As an example, suppose a file API has an error accessing a file. We might check that the error message contains the filename. Or we might check that the error message contains the path and filename if the directory portion is likely to be relevant.

As a final thought, if we want to allow for parallel unit tests on multiple threads, the (global) `logger` variable could be made into a thread local variable. That way each unit test, running on it's own thread, could have it's own mock logger object, allowing the unit tests to run in parallel without stepping on each other. This might be a good idea anyway, as logging from multiple threads would either need separate loggers, or the logger would have to be multithread aware and provide for proper safety or synchronization.

So yes, we should morph things into better tests for decent error messages. :)

DanRStevens:
Correct about the `__LINE__` reporting becoming less granular. I had paused to think about that while making the changes. There didn't seem to be an easy way to preserve the line numbers, at least nothing short of a custom exception class that stores the extra fields. The code flow in that area is pretty easy to follow though, and it's all still reported in the same file, which isn't all that big. It also occurred to me the error messages themselves were quite searchable.

----

I considered writing unit tests for `LocalResource`. That's always a good idea when adding new code. After sitting and thinking about it for a long while, I couldn't think of any meaningful tests I could write. In fact, after much Googling, I still have no idea how you'd write a meaningful unit test for pretty much any RAII like object.

The one restriction on the class, is that it's designed to work with pointers to a local resource. I added a `static_assert` to ensure that was the case. That's a compile time check though, which leads to a compile error, so there's no way to write a runtime unit test for that. Previously, I tried to check for it statically by peeling the pointer off, so that type deduction would enforce use with a pointer, though I found that didn't play so well with `LPSTR` like types if you were being explicit about the type parameter, hence the change to the `static_assert`.

I also thought about the public surface area of the class, but it's pretty minimal. It can be constructed, destructed, and converted automatically to it's underlying type. There are no particularly meaningful methods or return values to check. At best I could check that it converts to the value it was initialized with. I'm not sure that really warrants a test.

The main point of the class is that it frees the resource upon destruction. The mechanics of how it works is tied in to the structure of C++ code, and is enforced by the compiler. Any tests here would be more like checking if the C++ compiler was properly standards conformant, than anything specific about the code. I wrote in some logging statements while developing it, to make sure it was calling the destructor at the right time, but again, this was pretty much just checking the compiler did what it says it will do in the language spec.

I also considered putting an exception into the destructor that could report being freed on a bad value, but quickly dismissed it as a terrible idea. RAII like objects are automatically destroyed during exception unwinding. However, if an exception is thrown thrown when you're already in exception handling code, the runtime calls `terminate()`. That would completely defeat the purpose of using the object for safe exception handling.

As far as I can tell, there is no after the fact way of checking if a pointer has been freed. I don't know of any memory allocator that let's you check this.

I'm feeling a bit stumped at present.

----

Now that I can compile and run the project on Linux, this change set is actually tested![1]  :grinning:  Every commit was checked to make sure the unit tests ran correctly, and anything remotely suspect was tested in game. Fingers crossed I tested everything I needed to.

[1] Apologies for previous behaviour.

DanRStevens:
You know, I just remembered [`unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) can take a custom `Deleter`, which allows it to be used as a general scope bounded RAII type object. I read about it before, but had completely forgotten about it. I can probably reduce the `LocalResource` class to just a `using` clause for `unique_ptr`. It may require a few syntactic updates though.

One caveat is that `unique_ptr` typically takes the base type with the pointer peeled off. I may need to think about the implications of that.

Still, it kind of sidesteps much of the unit test question, by using the standard library, rather than custom code.

Oh, and I learned about something really neat. Many allocator aware objects, such as `unique_ptr` don't necessarily need to store a pointer to the `Deleter`, if defined right, due to the [Empty base optimization](https://en.cppreference.com/w/cpp/language/ebo).

DanRStevens:
Thanks for the review. This one turned out to be a bit large. And your question forced some very educational reading on me. :grin:

PR #127, opened 2019-09-19
Update windows error reporting

DanRStevens:
This cleans up the `GetLastErrorString` function, and improves the interface slightly.

----

Potential future work:

There is still one use of a local buffer that must be freed with `LocalFree`, which is not handled in an exception safe manner. If the `std::string` methods throw, perhaps due to memory exhaustion, we would be leaking that memory. In PR #126, there was a `LocalResource` class that was added to handle such resources in an exception safe way using RAII. We could potentially use it here, though it seems it's going to need a little bit of work to add a few extra needed features.

The calls to `GetLastErrorString` could potentially do more of the string formatting. We probably don't actually need to pass the `functionName` parameter. That part of the formatting could be handled by the caller. It's also questionable if we really need the error code number in the error message. Removing both could simplify the method and make it much more highly targeted. Would love some feedback on that idea.

### DIFF
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -94,7 +94,7 @@ void ConsoleModuleLoader::LoadModuleDll()
 	}
 	else {
 		const std::string errorMessage("Unable to load console module's dll from " + dllName +
-			". " + GetLastErrorStdString("LoadLibrary"));
+			". " + GetLastErrorString("LoadLibrary"));
 
 		PostErrorMessage(__FILE__, __LINE__, errorMessage);
 	}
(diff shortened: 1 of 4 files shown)


DanRStevens:
Eh, the function name argument was part of the original sample code that method was based on. Took me a while to even realize it wasn't that important, just part of the formatting. I initially though the function name might some sort of relevant to retrieving the actual error message. Kind of hard to know when using somewhat obscure Win API functions.

If we want to remove the parameter, we can do so in a future branch. I'll create an issue for it.

As for the "%1", I think that's related to the `FORMAT_MESSAGE_IGNORE_INSERTS` option. I suppose we could play with it if we wanted to.

PR #128, opened 2019-09-20
Make `GetLastErrorString` exception safe

DanRStevens:
This branch only really has 1 new commit. It pulls in the changes from PR #126 and PR #127 (see the two non-fast-forward merge commits at the end), and then applies one small fix on top of that.

Not sure if there is an easier way to present this for review, other than to have waited until the other two branches are merged to master, and then rebase the work on top of that.


DanRStevens:
Had a merge conflict to clean up here. Found it was easier to just rebase. That also cleans up some of the noise here, now that the other two dependent branches have been merged.

PR #132, opened 2019-09-20
Add Linux support for TestModule

DanRStevens:
This is mostly a Linux only change, though I did add a missing header file.


TestModule/TestFunctions.cpp
```diff
@@ -5,6 +5,8 @@
 #include <windows.h>
 #include <cstddef>
 #include <vector>
+#include <algorithm>
```
Brett208: Oh, I missed this, thanks for catching it.

Issue #121, opened 2019-09-11
Console Module Resource Loading Broken


DanRStevens:
As discovered in #110. It seems the ConsoleModLoader used to be able to load VOL files from the module directory, which would override built in VOL files. Currently this is broken.

The ConsoleModLoader should still be able to load loose files from the module folder, which will override the built in defaults. That part continues to work.

A possible solution might be to add `LocateVolFiles(consoleModLoader.GetModuleName())` in `ExtInit` in DllMain.cpp before it adds other VOL files, such as in the "Addon" folder. That should restore loading of VOL files form the console module folder, and have them take priority over the built in ones.

I believe this broke back when custom VOL lists were implemented. Previously VOL files were loaded by the game in `ResManager.Init()`, using `ResManager.GetFilePath`. The op2ext library changed this to search for VOL files itself, and load them directly, bypassing the internal mechanism, and the ability to override it with the `ART_PATH` setting or by later hooking of `ResManager.GetFilePath`.

Brett208:
This should be a priority fix before the next formal release. It may be broken in the last release of op2ext.

DanRStevens:
It looks like the bug presented here has been resolved satisfactorily.

Before closing this, should we add corresponding code for INI loaded modules? That would make module use much more symmetric. Then a module could be loaded either through the console, or through the INI file, and it would load resources along with it. We could also push this to a new issue. Loading of non-packed files would also need to be looked at.

Brett208:
I would prefer opening a new issue although I support the idea of ini files loading resources from their local directory.

DanRStevens:
Ok, new issue opened for tracking the corresponding update for INI modules. Thank you for the feedback.

PR #134, opened 2019-09-21
Load VOL files from console module folder

DanRStevens:
This is one possible way to address Issue #121.

I have not been able to test this change. It seems any use of `/loadmod` with a valid module is affected by the Mingw code generation bug.

----

One caveat about this change, is it doesn't work exactly the way console module VOL file loading used to work before. The difference is a bit subtle, and probably not significant to any existing modules.

Previously, the VOL loader used a fixed list of VOL files, which would come from the main game folder. When a console module was loaded, the console module first was searched first, before the main game folder. In either case, it was the same fixed list of named VOL files that was loaded. This would allow console module VOL files to completely replace the built in ones of the same name. If a file was missing from the new replacement VOL file, it would effectively be deleted from the game.

The new code loads any VOL file found in the the console folder, as well as any VOL file found in the main game folder. There is no longer a fixed list. If a VOL files of the same name exist in both folders, they will both be loaded, with the console module VOL file taking priority. As both modules are now loaded, and have a priority order, it is no longer possible to delete a file from the game by providing a replacement VOL file with that file missing. The file will simply not be seen in the replacement VOL file, and search will continue into the regular built in VOL files, where it will still be found.

If this is close enough behaviour, then this should be able to close the issue.

----

Potential future work:
We may want to also loop over the INI module folders, and load any VOL files found in them too.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -75,6 +75,8 @@ int TApp::Init()
 	// Load all active modules from the .ini file
 	iniModuleLoader.LoadModules();
 
+	// ConsoleModule name matches relative folder from game exeucutable folder
+	LocateVolFiles(consoleModLoader.GetModuleName());
 	LocateVolFiles("Addon");
 	LocateVolFiles(); //Searches root directory
 


Brett208:
I think this breaking change for ConsoleModules is acceptable. I cannot think of a past or future use case where you would want to delete an existing file from loading by excluding it from your module's version of a volume. However, the documentation should be updated to reflect.

I should be able to test this code in the evening today.

PR #135, opened 2019-09-21
Lift `UNICODE` restriction on string conversion

DanRStevens:
Previously the code tried to issue a runtime exception for what was effectively a compile time error. It also provided a misleading function name in that the "T" in "LPCTSTR" meant it would adapt to changes in string width based on the `UNICODE` setting. The function however only actually provided support for narrow strings, not for wide strings.

By providing two overloads of the function, for both types of strings, the compiler will auto detect at compile time which function should be called. It also means both versions of the method could be used in the same executable, allowing support of mixed string width code. If the wrong string types are mixed together in an incompatible way, it will result in a compile time error due to the type mismatch. And finally, it means the function will finally live up to the "T" in its name.

### DIFF
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -3,11 +3,11 @@
 
 std::string ConvertLpctstrToString(LPCSTR str)
 {
-#if defined(UNICODE) || defined(_UNICODE)
-	throw std::runtime_error("Unicode is not supported");
-#endif
-
 	return std::string(str);
+}
+std::wstring ConvertLpctstrToString(LPCWSTR str)
+{
+	return std::wstring(str);
 }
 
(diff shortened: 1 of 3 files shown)


Brett208:
I'd like to review and test this branch, but do not have time today. If it is taking too long and is in the way of other work, go ahead and merge. Otherwise I'll try to get to it in the next 2 days.

DanRStevens:
Err, wait, unable to load the shell DLL? Or unable to compile?

I tried compiling with `-DUNICODE` (define `UNICODE`), and it failed to compile due to type errors. Which made sense. Some method calls weren't designed to handle either.

Also, looking at these errors makes me realize some of the old code was a bit sloppy in its handling of the matter. I'll open a new issue to track.

Issue #131, opened 2019-09-20
Refactor and add support for better error log message testing


DanRStevens:
This might be a good problem to try Google Mock on. If we can create a mock object for the logger, than we can verify more directly if we are logging proper error messages, without needing to actually go to disk.

To test with mock objects (with or without Google Mock), we'd need to make a few changes to the logging code. We'd have to split the `Logger` class into an interface class, and a separate implementation. This might be a good idea to do anyway, since it would allow for multiple logging output options.

We would also need to change the global `logger` variable to be a pointer or reference to a `Logger`. That would allow method calls to go through to a potential derived class. I'd recommend a pointer if we want to allow for updates to the logging output at runtime, allowing it to be swapped out for a different type of logger. A global reference can only be set once, and it would be done during global initialization.

Not required, but something I was thinking about. Perhaps we should use global logging functions, rather than a global logging object. The global functions can then delegate to the logging object internally. That would allow restricting access to the logging object, and possibly making it not so global. That could have benefits for deferring the creation of the logging object. It would also protect the code from having to change `logger.Log` to `logger->Log` all over the place if we change it from a `Logger` instance to a pointer.

Finally, with Google Mock / Google Test support, we might potentially use the string matchers on the error messages. That would allow for testing of keywords, rather than exact error messages. Small changes in the error messages wouldn't immediately break tests, and we also wouldn't have to export the error strings to give tests access. As an example, suppose a file API has an error accessing a file. We might check that the error message contains the filename. Or we might check that the error message contains the path and filename if the directory portion is likely to be relevant.

As a final thought, if we want to allow for parallel unit tests on multiple threads, the (global) `logger` variable/pointer could be made into a thread local variable. That way each unit test, running on it's own thread, could have it's own mock logger object, allowing the unit tests to run in parallel without stepping on each other. This might be a good idea anyway, as logging from multiple threads would either need separate loggers, or the logger would have to be multithread aware and provide for proper safety or synchronization.

Brett208:
I think you are discussing using the Singleton pattern for our logger (https://en.wikipedia.org/wiki/Singleton_pattern). I think it is an acceptable fit.

You also mentioned switching to using an external logger. I'd be happy to try that out as well. If we define our own interface, we could use an external logger without letting it be tightly coupled to our own code base. Maybe best to write the interface first. 

This is all a lot of work though, and maybe better on the backburner for now? Maybe what I'm getting at is that I don't plan to make this a priority but would welcome the changes.

[...]

PR #141, opened 2019-09-21
Logger updates

DanRStevens:
Here is some refactoring to help support Issue #131.

I was starting on some Google Mock tests, but it seems there's a version mismatch between the documentation I was reading, and what I have installed. I think I'll punt on that part for now.

The changes introduce a new global `Log(message, module = "op2ext")` function. This function then delegates to the `Logger`. Access to the former global variable is now much more restricted. The `Logger` was split into an interface and a `LoggerFile` implementation.

The new global `LoggerFile` object has been moved to DllMain.cpp, rather than being in the static library. This is a much more sensible place for it, and interferes less with testing. The static library should just be providing methods that can be called, not dictating global objects that must exist. I plan to create an issue about moving the other globals.

I carefully checked the code and C++ reference documents on initialization order to make sure the logging code didn't violate anything. Hopefully I got it all right.

I didn't bother with thread local variables for this update. We can look into that in a future update.

### DIFF
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -1,9 +1,8 @@
 #include "Log.h"
-#include "LoggerFile.h"
 #include "op2ext-Internal.h"
 
 
 void Log(const std::string& message, const std::string& moduleName) {
 	// Delegate to internal logger
-	logger.Log(message, moduleName);
+	logger->Log(message, moduleName);
 }
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -1,0 +1,9 @@
+#include "Log.h"
+#include "Logger.h"
+#include "op2ext-Internal.h"
+
+
+void Log(const std::string& message, const std::string& moduleName) {
+	// Delegate to internal logger
+	logger.Log(message, moduleName);
+}
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -1,5 +1,5 @@
 #include "Log.h"
-#include "Logger.h"
+#include "LoggerFile.h"
 #include "op2ext-Internal.h"
 
 
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -2,6 +2,13 @@
 #include "op2ext-Internal.h"
 
 
+// Set logging output destination
+// Caller owns the logger and is responsible for cleanup when logger is no longer required
+// Use `SetLogger(nullptr);` to unset a logger
+void SetLogger(Logger* newLogger) {
+	logger = newLogger;
+}
+
 void Log(const std::string& message, const std::string& moduleName) {
 	// Delegate to internal logger
 	logger->Log(message, moduleName);
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -10,6 +10,9 @@ void SetLogger(Logger* newLogger) {
 }
 
 void Log(const std::string& message, const std::string& moduleName) {
-	// Delegate to internal logger
-	logger->Log(message, moduleName);
+	// Make sure a logger has been set first
+	if (logger) {
+		// Delegate to internal logger
+		logger->Log(message, moduleName);
+	}
 }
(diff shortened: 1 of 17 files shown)


srcStatic/Log.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "op2ext-Internal.h"
+
+
+// Use anonymous namespace to keep globals private
+// Prevents other files from gaining access with a sneaky extern declaration
+namespace {
+  // Static (constant and zero) initialization happens before dynamic initialization
+  // Such values are well defined if accessed by a global object constructor before Main/DllMain starts
+  // This pointer is safely set to null before other globals are initialized
+  Logger* logger = nullptr;
```
Brett208: Why did you choose to use a 'raw' pointer here instead of a unique or shared pointer?

Should we ensure the logger is deleted before closing out of the application?

Thanks,
Brett

DanRStevens: I documented that responsibility was for the caller. Hence no ownership transfer takes place. This makes it safe to pass local or global objects. I used this is my GMock attempt, before I realized I had an older version installed which didn't work with the syntax I wanted. The test case would have looked like:
```
TEST(Log, SetLoggerReceivesMessages) {
	const auto message = std::string("Logger should receive this");

	MockLogger logger;
	EXPECT_CALL(logger, Log(message));

	EXPECT_NO_THROW(SetLogger(&logger));
	EXPECT_NO_THROW(Log(message));
	EXPECT_NO_THROW(SetLogger(nullptr));
}
```

The `logger` instance is local and would have been cleaned up automatically when the test function ended. If I'd put `unique_ptr` into the method signature, there would have been no safe way to use a local or global object. I would have been forced to wrap it in a `unique_ptr`, which would have forced me to use dynamic allocation.

The caller is of course free to use `unique_ptr` if they want.
```
unique_ptr<Logger> logger;

void CreateLogger() {
  logger = make_unique<LoggerFile>();
  SetLogger(logger.Get());
} // No automatic cleanup
```

In this case, when the application ends, the runtime will ensure cleanup of the `Logger` object stored in the global `unique_ptr`.

I could also have use a global:
```
LoggerFile logger;

void Init() {
  SetLogger(&logger);
  // ...
}
```

Again, this object is automatically destructed when the application terminates.

----

By keeping the `unique_ptr` out of the interface it gives the user the choice of how they want to manage object lifetime, and gives them the flexibility to use objects that are either dynamically allocated, allocated on the stack, or global.

DanRStevens:
Yes, I would like to add such a test case. Right now, I could ensure it was safe to call `Log` with a `LoggerFile` instance. However it's more awkward to ensure that it actually works when using that logging destination.

It was for this test case and this very reason that I was trying to use Google Mock. Perhaps I can get it working today using the older syntax.

DanRStevens:
Hmm, seems there's no Google Mock package installed for the AppVeyor builds.

@Brett, was there a handy GMock package that could be installed? The tests pass on Linux, so it should just be a matter of installing an appropriate GMock library.

If that can't be done easily, maybe we can peel off the GMock stuff into a different branch.

Brett208:
Locally I switched from nuget to vcpkg. In release mode everything compiles fine. In debug mode, I get the following assertion:

debug assertion

I haven't figured out what is causing it or if we can somehow ignore it.

Anyways, if that is solved, we can switch to vcpkg and have gmock. The official Microsoft nuget package for gtest does not contain gmock for some reason.

I'd be happy if you wrap the test somehow so it only runs on Linux/mingw builds, merge and open an issue for MSVC gmock support.

Thanks, Brett

DanRStevens:
I had an idea. I added a `__has_include` guard around the GMock related test code. That way if you have GMock installed, through any means, be it NuGet, vcpkg, Apt (Linux), or some custom local install, the test will compile and run. If not, it's skipped. That should provide more coverage and flexibility than a platform/compiler detect.

I'm tempted to wrap it in a platform detect anyway, based on your debug assertion. Though I'm also a little uncertain if the debug assertion is coming from the GMock related code. It doesn't really give enough detail to be certain. One possibility is if GMock is somehow dependent on Debug/Release settings, and you only have a Release version of GMock installed. It could also be there is a pointer or memory management bug in the code changes.

Maybe I should give [Valgrind](http://www.valgrind.org/) a try and see if it reports anything. I've heard of Valgrind many times, but have never really tried it out before.

PR #144, opened 2019-09-22
Move `StringSplit` method

DanRStevens:
The `SplitString` and `TrimString` methods are better declared among the other string functions.

I added some unit tests. It looks like we may have a few corner cases that may not behave as expected. I left the offending checks commented out. I don't think we need to fix them at this time. They don't seem to be particularly relevant.

I also noted we may want to avoid designs where functions take an `enum` parameter, and that parameter is likely to be a fixed constant at the call site. I'll discuss it further in an issue, along with a few alternatives. Any bug fixes (assuming we agree they are bugs) can be rolled into that.

### DIFF
--- a/srcStatic/GlobalDefines.cpp
+++ b/srcStatic/GlobalDefines.cpp
@@ -3,8 +3,6 @@
 #include "op2ext-Internal.h"
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
-#include <sstream>
-#include <cstddef>
 
 bool modalDialogsDisabled = false;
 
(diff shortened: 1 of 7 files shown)


Brett208:
I won't have time to review this code today. Please feel free to merge unless you really want me to load, test and review closely. It looks like it is generally doing better at running unit tests against it.

DanRStevens:
Alright, I don't think this one needs to be reviewed too closely. It's mostly just moving code around. About the only potentially active change is adding defaults to a couple of parameters. None of the function code has been changed. I'll just merge for now. Feel free to review later if you feel like it. Or not.

Issue #149, opened 2019-09-24
Review `UNICODE` handling


DanRStevens:
As noted in PR #135, some of the older existing code is a bit sloppy with regards to `UNICODE` aware functions. In a few places we may have used the generic type, which is able to change string width, in places where a fixed type should be used.

For instance, the `LoadLibrary` call from Outpost2.exe which handles loading the shell DLL has a fixed width. It was baked into Outpost2.exe when it was compiled. Our replacement method uses the generic width names. Changing the `UNICODE` setting in our project shouldn't affect (and doesn't affect) the point in Outpost2.exe which is hooked.

----

On Linux, as quick way to check for errors is to add the following to the makefile:
```
CPPFLAGS := -DUNICODE
```

DanRStevens:
From Raymond Chen's blog:
[TEXT vs. _TEXT vs. _T, and UNICODE vs. _UNICODE](https://devblogs.microsoft.com/oldnewthing/20040212-00/?p=40643)

DanRStevens:
While investigating I encountered one strange problem. The [`MODULEENTRY32`](https://docs.microsoft.com/en-ca/windows/win32/api/tlhelp32/ns-tlhelp32-moduleentry32) struct has a `szModule` field, which is defined as `char`. The related struct [`MODULEENTRY32W`](https://docs.microsoft.com/en-ca/windows/win32/api/tlhelp32/ns-tlhelp32-moduleentry32w) has the same `szModule` field, but is defined as `WCHAR`.

It seems that defining `UNICODE` will map the name `MODULEENTRY32` to `MODULEENTRY32W`. There is no corresponding `MODULEENTRY32A`. I'm not sure if this is a Windows API thing, or a Mingw thing. There's no documentation about it. Similarly, there is no `Module32FirstA` nor any `Module32NextA` function, only the plain or "W" variant.

It would be possible to reference the narrow string version of the struct by using the tag name directly: `tagMODULEENTRY32`. Unfortunately without also using corresponding non-wide versions of the functions, you'll still get type errors. I think I'll leave this one for now.

----

A lot of the issues I looked at were resolved in favour of just using narrow strings everywhere, and the "A" versions of Windows API methods. We could potentially lift some of our narrow string assumptions, though I don't think that's a high priority item.

[...]

PR #150, opened 2019-09-24
Unicode fixes

DanRStevens:
This resolves some of the earlier sloppiness regarding the `UNICODE` setting from Issue #149. It doesn't fix everything, but it should resolve most problems.

Mostly, I explicitly resolved in favour of narrow strings, since much of the code base has such an assumption built in. We can revisit this later if we ever decide to put effort into relaxing those assumptions.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -15,7 +15,7 @@
 void LocateVolFiles(const std::string& relativeDirectory = "");
 
 // Declaration for patch to LoadLibrary, where it loads OP2Shell.dll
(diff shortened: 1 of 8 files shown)


DanRStevens:
Err, the commit comment should actually have referenced Issue #149. After writing up the details I was suddenly hit with a way to get a `UNICODE` build to work. I wouldn't say we properly support Unicode by any means, just that it should compile and run without error. Mostly because a number of method calls are now explicitly resolved to narrow string variants, regardless of the Unicode setting.

DanRStevens:
Interesting. I didn't see that in my testing on Linux. Perhaps we can look into it sometime in the future.

PR #151, opened 2019-09-24
Split `TrimString`

### LINKED ISSUE #145: Remove `enum` from `SplitString` design
In PR #144 I noted the `SplitString` method takes an `enum` parameter which is likely to be a hardcoded constant at the call site. I believe `enum` parameters like this are often a suboptimal design and would like to suggest a few alternatives.

----

The first alternative is to remove the `enum` parameter and provide separate methods for each of the `enum` values. The programmer then selects the behaviour by function name, rather than by hardcoded parameter. In this example, that might look something like this:
```
// Old method
std::vector<std::string> SplitString(std::string stringToSplit, char delimiter, TrimOption trimOption = TrimOption::Both);

// New methods
std::vector<std::string> SplitString(std::string stringToSplit, char delimiter);
std::vector<std::string> SplitStringTrimLeading(std::string stringToSplit, char delimiter);
std::vector<std::string> SplitStringTrimTrailing(std::string stringToSplit, char delimiter);
std::vector<std::string> SplitStringTrimBoth(std::string stringToSplit, char delimiter);
```

This is only feasible when the number of `enum` constants is relatively few.

There is however a complication in this particular case. Here the `enum` parameter is passed to `SplitString` but not actually used until the internal call to `TrimString`. If we took this idea to it's natural conclusion, we'd also have to split `TrimString`. We would then have 8 new methods instead of 2 old methods. Worse, if there are multiple `enum` parameters, we may end up with a combinatorial explosion in the number of functions. For that reason, using the `enum` parameter can be considered fairly practical, particularly for plain C code.

----

A second option, available to C++, is to use templates. The `SplitString` method could become a template method, where the template parameter is the `TrimString` method. Assume the `TrimString` method is split as above into 4 distinct methods. We can now produce 4 types of `SplitString` methods using only a single template function definition. Further, we could set a default template parameter if one case is particularly common. Calling code might look something like:
```
auto items = SplitString(stringToSplit); // Default to no trimming
auto trimLeadingItems = SplitString<TrimLeading>(stringToSplit);
auto trimTrailingItems = SplitString<TrimTrailing>(stringToSplit);
auto trimBothItems = SplitString<TrimBoth>(stringToSplit);
```

----

A third option, available to functional programming environments, is to not make one method dependent on the other, but rather to combine them with chained calls, using a ["collect" (aka "map")](https://en.wikipedia.org/wiki/Map_(higher-order_function)) to do the array processing. In a functional language like Ruby, that might look something like:
```
items = stringToSplit.split(',').map{|x| x.trim }
```

Roughly, the above code first does the split, returning an array, and then makes a chained `map` call on the array, which loops over the array elements and calls `trim` on each element. The result in an array of trimmed values.

The rough C++ equivalent of the `map` call is [`std::transform`](https://en.cppreference.com/w/cpp/algorithm/transform). It loops over an arbitrary container, and calls a given method on each item. Unfortunately it's a bit verbose to use at present, due to 3 iterator parameters. When C++ adds Ranges, it should clean up some of the calling syntax noise. Worse though, is it's not chainable. It's not a method on `vector` so you can't use `someVector.transform(...)`, and it doesn't return a value, so you can't capture the result with a one liner: `auto someVar = someVector.transform(...)`.

It seems C++ still has a bit of a ways to go for functional programming.

For now, I think it needs to be a two liner:
```
auto result = SplitString(stringToSplit);
std::transform(result.begin(), result.end(), result.begin(), &Trim);
```

----

I propose we do the following:

Split out `TrimString` into 3 non-trivial options: `Trim` (both), `TrimFront`, `TrimBack`.

Make the `SplitString` and `TrimString` methods independent from each other. The `SplitString` method will no longer be concerned about trimming. To perform both, the operations will need to be chained independently.

Use a multi-line solution to perform both operations. As the current `SplitString` method is only called from a one place in `ModuleLoader::GetModuleNames`, this shouldn't be very onerous, and the details will still be hidden away inside a method, albeit one that's specific to splitting and trimming modules names from an INI file.

DanRStevens:
This closes #145.

I ended up using a bit of a hybrid approach, using aspects from all 3 suggested styles.

It occurred to me the `ModuleLoader::GetModuleNames` method was untested. To avoid dumping new code with significant changes into that method, I choose to keep a `SplitAndTrim` method within the StringConversion library code. It delegates to the `Split` method and the `Trim` method, where the `Trim` method is actually a template parameter, with a default to trim both ends. All 3 types of methods can be called independently, and all have their own test code. That minimized changes needed for the calling code.

### DIFF
--- a/srcStatic/ModuleLoader.cpp
+++ b/srcStatic/ModuleLoader.cpp
@@ -47,7 +47,7 @@ bool ModuleLoader::IsBuiltInModuleRequested(const std::string& moduleName)
 std::vector<std::string> ModuleLoader::GetModuleNames(const std::string& moduleType)
 {
 	std::string sectionNames = GetOutpost2IniSetting("Game", moduleType);
-	std::vector<std::string> sectionNamesSplit = SplitString(sectionNames, ',', TrimOption::Both);
+	std::vector<std::string> sectionNamesSplit = SplitAndTrim(sectionNames, ',');
 
 	return sectionNamesSplit;
 }
--- a/srcStatic/ModuleLoader.cpp
+++ b/srcStatic/ModuleLoader.cpp
@@ -46,10 +46,10 @@ bool ModuleLoader::IsBuiltInModuleRequested(const std::string& moduleName)
 
 std::vector<std::string> ModuleLoader::GetModuleNames(const std::string& moduleType)
 {
-	auto sectionNames = GetOutpost2IniSetting("Game", moduleType);
-	auto sectionNamesSplit = SplitAndTrim(sectionNames, ',');
+	auto sectionNamesString = GetOutpost2IniSetting("Game", moduleType);
+	auto sectionNames = SplitAndTrim(sectionNamesString, ',');
 
-	return sectionNamesSplit;
+	return sectionNames;
 }
 
 std::string ModuleLoader::GetModuleName(std::size_t index)
--- a/srcStatic/ModuleLoader.cpp
+++ b/srcStatic/ModuleLoader.cpp
@@ -46,8 +46,8 @@ bool ModuleLoader::IsBuiltInModuleRequested(const std::string& moduleName)
 
 std::vector<std::string> ModuleLoader::GetModuleNames(const std::string& moduleType)
 {
-	std::string sectionNames = GetOutpost2IniSetting("Game", moduleType);
-	std::vector<std::string> sectionNamesSplit = SplitAndTrim(sectionNames, ',');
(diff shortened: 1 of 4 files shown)


DanRStevens:
I was seriously tempted to further simplify the calling code to:
```
std::vector<std::string> ModuleLoader::GetModuleNames(const std::string& moduleType)
{
	return SplitAndTrim(GetOutpost2IniSetting("Game", moduleType), ',');
}
```

But then decided against that on readability grounds.

PR #154, opened 2019-09-24
Move globals

### LINKED ISSUE #142: Move globals from static library to DLL
The DLL would make for a much better home for the majority of the global variables. At this point that's where most of them are used anyway.

The following globals are only used from the DLL:
```
bool modulesRunning = false;
VolList volList;
ConsoleModuleLoader consoleModLoader(FindModuleDirectory());
ModuleLoader moduleLoader;
```

The following global is applicable to the static library:
```
std::string moduleDirectory; // Must be defined + initialized before consoleModLoader
```

I believe the `moduleDirectory` variable can be made a static variable in a private static method of the `ConsoleModuleLoader` class. That should restrict access so it's no longer global, and avoid the static initialization order problem, since function level static variables are instantiated on first use. That ensures a well defined initialization time, which is before it actually gets used for something.

The static library should include mainly functions and class, rather than global variables. This should also help simplify issues with unit testing.

DanRStevens:
This closes #142.

The `moduleDirectory` variable was moved into the `ConsoleModuleLoader` class as a function level static of a private static function.

The remaining globals were moved to the DLL project.

Some header cleanup was done.


srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -173,3 +180,12 @@ void ConsoleModuleLoader::RunModule()
 		}
 	}
 }
+
+std::string& ConsoleModuleLoader::ModuleDirectory()
+{
+	// Function level statics are initialized on first use
+	// They are not subject to unsequenced initialization order of globals
+	// There is no order dependency on when this variable may be used
+	static std::string moduleDirectory;
```
Brett208: Is there a way to mark this as const, so the moduleDirectory cannot be changed after initialization?

DanRStevens:
I was initially hoping to mark it as `const` or something, until I realized that would be impossible.

For one, the module path is set at runtime, so it can't really be a `const`. The only way to make it a `const` is to make it a member of a class with delayed construction. But that's just pushing the construction/initialization problem around, not solving it.

The problem is there is no way to initialize the value from within the function. To make the function easy to use as an accessor, there are no parameters to the function. As the variable is a function level static variable, and we are trying to avoid passing information around in globals, the only way left to communicate is through the parameters and the return value. But if there are no parameters, there's also no way to set the value inside the function, other than to give access through the return value.

There's also no way to write another function to mutate it, since another function wouldn't have access to the original function's static variables.

As for controlling access, keep in mind the static method is private. Hence the only way to get the writable reference to that variable is from within the class.

DanRStevens:
Ahh yes, maybe I should have documented that. that was a bit of a subtle point.

PR #156, opened 2019-09-25
Update `LocalResource`

### LINKED ISSUE #143: Update `LocalResource` to use `unique_ptr` with custom `Deleter`
This relates to PR #126 and PR #128, which introduced `LocalResource` and used it.

It was later discovered in PR #126 that [`unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) with a custom `Deleter` could be used to implement `LocalResource`. The implementation might look similar to:
```
struct LocalDeleter {
	void operator()(void* resource) { LocalFree(resource); }
};

template <typename T>
using LocalResource = std::unique_ptr<T, LocalDeleter>;
```

Further, it was noted that due to the [Empty Base Optimization](https://en.cppreference.com/w/cpp/language/ebo) there would be no runtime storage needed for a pointer to the `Deleter`.

This should perhaps be investigated and used to simplify the code. Perhaps once both PRs have been merged to master, the class can be updated. That way there's no need to coordinate changes across branches.

DanRStevens:
This closes #143.

First a bit of cleanup and generalization of `LocalResource` so the final change can be seen easier. The actual change to `unique_ptr` will come in a later push.

Edit: Manage to solve the one thing I didn't quite like about the `unique_ptr` solution. I think this is good now.

### DIFF
--- a/srcStatic/GetCommandLineArguments.cpp
+++ b/srcStatic/GetCommandLineArguments.cpp
@@ -9,7 +9,7 @@ std::vector<std::string> GetCommandLineArguments()
 {
 	std::vector<std::string> arguments;
 	int argumentCount;
-	LocalResource<LPWSTR*> commandLineArgs = CommandLineToArgvW(GetCommandLineW(), &argumentCount);
+	LocalResource<LPWSTR[]> commandLineArgs = CommandLineToArgvW(GetCommandLineW(), &argumentCount);
 
 	if (commandLineArgs == nullptr) {
 		throw std::runtime_error("Unable to retrieve command line arguments attached to Outpost2.exe.");
--- a/srcStatic/GetCommandLineArguments.cpp
+++ b/srcStatic/GetCommandLineArguments.cpp
@@ -9,7 +9,7 @@ std::vector<std::string> GetCommandLineArguments()
 {
 	std::vector<std::string> arguments;
 	int argumentCount;
-	LocalResource<LPWSTR[]> commandLineArgs = CommandLineToArgvW(GetCommandLineW(), &argumentCount);
+	LocalResource<LPWSTR[]> commandLineArgs{CommandLineToArgvW(GetCommandLineW(), &argumentCount)};
 
 	if (commandLineArgs == nullptr) {
 		throw std::runtime_error("Unable to retrieve command line arguments attached to Outpost2.exe.");
--- a/srcStatic/LocalResource.h
+++ b/srcStatic/LocalResource.h
@@ -8,7 +8,8 @@
 // RAII management of LocalAlloc/LocalFree resources
 template <typename T>
 class LocalResource {
-	static_assert(std::is_pointer_v<T>, "Type must be a pointer");
+	using PointerType = T;
+	static_assert(std::is_pointer_v<PointerType>, "Type must be a pointer");
 public:
 	LocalResource() : resource(nullptr) {
 	}
@@ -20,9 +21,9 @@ public:
 	}
 
 	// Auto convert to underlying type
-	operator T() const {
+	operator PointerType() const {
 		return resource;
 	}
 private:
-	T resource;
+	PointerType resource;
 };
--- a/srcStatic/LocalResource.h
+++ b/srcStatic/LocalResource.h
@@ -10,7 +10,7 @@ template <typename T>
 class LocalResource {
 	static_assert(std::is_pointer<T>::value, "Type must be a pointer");
 public:
-	LocalResource() : resource(NULL) {
+	LocalResource() : resource(nullptr) {
 	}
 	LocalResource(T resource) : resource(resource) {
 	}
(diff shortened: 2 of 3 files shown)


DanRStevens:
I realize in hindsight that the `const` fix only works for a single level of pointers, when really I think it should ideally work for all levels of pointers. I'm looking into it now. Not really a show stopper, just a nice to have.

DanRStevens:
Nope, my mistake. The `const_cast` in the `Deleter` only needs to work on a single level of pointers. Anything below the first level is irrelevant when converting to `void*`. There's no need to adjust that part.

----

I was seeing an error with a `unique_ptr` construction when I tried adding `const` to the data type. This was the real error, which I was misdiagnosing. The line in question was:
```
LocalResource<LPWSTR[]> commandLineArgs{CommandLineToArgvW(GetCommandLineW(), &argumentCount)};
```

Now the declared return type of `CommandLineToArgvW` is `LPWSTR*`. As we were not going to modify the result, I thought I'd try storing it in a `const` variable. After all, pointers to a type should convert automatically to pointers to a `const` of that type.

When I tried changing the `LPWSTR[]` to `LPCWSTR[]` it produced an error. Turns out when you add `const` to a multi level pointer, you must also add it at intermediate levels as well. See the section titled "Qualification conversions" in [Implicit Conversions](https://en.cppreference.com/w/cpp/language/implicit_conversion) for details. The correct change was to `LPCWSTR const []`. Given my original intention this was also a better change. This change worked with the custom `LocalResource` code.

With the new `unique_ptr` based code, that change alone doesn't compile. The problem seems to be the `unique_ptr` constructor forbids implicit conversions. When constructing the object, the passed pointer must be the exact type the object expects. You can work around it with a `const_cast` to pre-convert, but that gets a bit ugly. It would look something like:
```
LocalResource<LPCWSTR const []> commandLineArgs{const_cast<LPCWSTR*>(CommandLineToArgvW(GetCommandLineW(), &argumentCount))};
```

----

Anyway, I'm not sure if we need this, or if we should just stick with the mutable pointer type which matches the function result. Or, we could use the original custom `LocalResource` which allows an implicit conversion. Would appreciate some feedback. The 3 options are:

1) Keep things mutable. (Current state of the code)
This syntax works with both the custom `LocalResource` implementation, and with the new current `unique_ptr` based code.
```
LocalResource<LPWSTR[]> commandLineArgs{CommandLineToArgvW(GetCommandLineW(), &argumentCount)};
```
2) Declare things as `const` with implicit conversion.
This syntax only works with the older custom `LocalResource` implementation. We would have to abandon the `unique_ptr` alias implementation.
```
LocalResource<LPCWSTR const []> commandLineArgs{CommandLineToArgvW(GetCommandLineW(), &argumentCount)};
```
3) Declare things as `const` with explicit conversion.
This syntax works with either the custom `LocalResource` implementation, and with the new current `unique_ptr` based code.
```
LocalResource<LPCWSTR const []> commandLineArgs{const_cast<LPCWSTR*>(CommandLineToArgvW(GetCommandLineW(), &argumentCount))};
```

Right now I'm inclined to go with option 1. It works now, it avoids a `const_cast`, and the `const` declaration for the read-only local variable doesn't seem like it would add much of significance to the code.

DanRStevens:
Is there anything I can do to make this one easier to review? It contains a number of advanced features that may be hard to read or reason about.

srcStatic/GetCommandLineArguments.cpp
```diff
@@ -9,7 +9,7 @@ std::vector<std::string> GetCommandLineArguments()
 {
 	std::vector<std::string> arguments;
 	int argumentCount;
-	LocalResource<LPWSTR*> commandLineArgs = CommandLineToArgvW(GetCommandLineW(), &argumentCount);
+	LocalResource<LPWSTR[]> commandLineArgs{CommandLineToArgvW(GetCommandLineW(), &argumentCount)};
```
Brett208: Sorry for the long wait on reviewing.

I don't understand exactly what is going on here with the curly braces and the brackets. 

It seemed more intuitive to me before this change as the function returned a pointer to an array of LPWSTR. Now I guess you are creating the array of pointers and storing it in the LocalResource? But the syntax isn't clear to me.

I'm sort of leaning towards leaving this as the raw pointer since I felt like I understood it better. But my lack of understanding isn't necessarily a good reason, plus you put time into updating it which I don't want to throw away...

DanRStevens: The `CommandLineToArgvW` method returns a dynamic array of strings. This is conceptually a `string[]`. As arrays decay to pointers in C++, we were able to store that into variable typed with a pointer rather than a dynamic array. That would effectively be a `string*`. As the explicit non-decayed array type is a bit more accurate, I chose to express that. Hence the change from `LPWSTR*` to `LPWSTR[]`.

The curly braces are uniform initialization syntax. It could potentially have just used regular parenthesis. It's just calling the constructor, and passing the argument inside the `{}`. It would be the same as: `commandLineArgs(CommandLineToArgvW(GetCommandLineW(), &argumentCount))`

PR #157, opened 2019-09-26
Cleanup `nullptr` checks

DanRStevens:
I temporarily set `-Wzero-as-null-pointer-constant` for the Linux build to scan for uses of the literal `0` with pointers. That led to the cleanup here. I skipped fixing one warning that is fixed in the other open PR #156.

I do not wish to enable this check fully at this time as the Google Test code contains many violations of the check. It generates way too much spam when the test projects are compiled with this flag.

I would be interested in setting just the core projects (op2extLib, op2extDll) to compile with the warning enabled. However, the current makefile design pulls in the core project settings when compiling the related test projects. I would have to enable some sort of bypass mechanism to set this warning for the core projects, but not for the associated test projects. That would be work for a different branch.

### DIFF
--- a/srcStatic/GameModules/IniModule.cpp
+++ b/srcStatic/GameModules/IniModule.cpp
@@ -39,7 +39,7 @@ HINSTANCE IniModule::LoadModuleDll()
 void IniModule::Load()
 {
 	// Call the InitMod function if it exists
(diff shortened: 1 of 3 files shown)


DanRStevens:
I had thought so too, but it turns out it was the NetFixClient that I had to take the setting from.

PR #158, opened 2019-09-26
Cleanup string wrapping

### LINKED ISSUE #148: Review Windows String code
Windows native raw strings are used in a few places in the code base. A few have been updated recently. We should perhaps take the time to look over the rest while it's still fresh. In particular, I'd like to update `ConvertLPWToString` from StringConversion.cpp, which hasn't been touched yet.

----

Quick summary:

Windows native raw strings are used in the following files:
srcStatic/StringConversion.h
srcStatic/StringConversion.cpp
srcStatic/GetCommandLineArguments.cpp
srcStatic/WindowsErrorCode.cpp
srcDLL/DllMain.cpp

Windows native raw strings form part of the public function signature for:
`ConvertLPWToString` - Converts Windows native raw string from wide to narrow, and then wraps in `std::string`
`ConvertLpctstrToString` - Wraps a Windows native raw string into `std::string`/`std::wstring` (changes/additions pending)

Windows native raw strings are processed internally by:
`GetCommandLineArguments` - Parsed command line arguments
`GetLastErrorString` - Win32 API error string
`LoadLibraryNew` - DLL name in event hook to tell when OP2Shell.dll is loaded

----

The `ConvertLPWToString` uses a `const LPWSTR pw` parameter. Windows provides the type `LPCWSTR` which expands to `const LPWSTR`, so we're being a bit inconsistent here.

----

The `<windows.h>` header is a big file, which can slow compilation, so we should try to minimize it's use when possible, particularly in header files. Using Windows specific types requires an include for `<windows.h>`. In particular, functions that use these types as part of their signature need to include `<windows.h>` in the header file for that function.

To reduce header pollution, functions that need Windows specific types in their signatures can be split into different files from other methods that don't use Windows specific types as part of their signature. This should reduce the transitive impact of header file includes.

Specifically, `ConvertLPWToString` and `ConvertLpctstrToString` can be moved to a separate file.

----

The `ConvertLPWToString` uses raw `new[]`/`delete[]` calls. We should switch that to either use `unique_ptr`, or perhaps a `std::string` or `std::vector` as the dynamic buffer. Using a `std::string` may even reduce the amount of copying needed to return a `std::string`. The would make the code cleaner, more robust to editing, and exception safe.

----

The signature of `ConvertLPWToString` returns a `std::string` through an out parameter and reports errors through a `bool` return value. We should perhaps use the more idiomatic C++ way, where we return the `std::string`, and report errors with exceptions.

----

The `ConvertLPWToString` method has effectively two responsibilities. One is to convert from wide strings to narrow strings. The other is to wrap the narrow string into an `std::string`. We already have the `ConvertLpctstrToString` method to wrap Windows raw strings into `std::string` (or `std::wstring`, change pending). Perhaps we should untie the two responsibilities, and re-use the other method.

In particular, a raw wide string could first be wrapped into an `std::wstring`, and then converted to a narrow `std::string`. The wrapped string buffers could potentially be passed directly to the [`WideCharToMultiByte`](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte) Windows API method.

----

There is an alternative way to convert between wide and narrow strings. Some C++ standard library methods can help here. Some relevant info can be found at:
[How to convert wstring into string?](https://stackoverflow.com/questions/4804298/how-to-convert-wstring-into-string)
One answer gives step by step code to convert wide strings to narrow, and then gives a one-liner for converting the opposite direction from narrow to wide:
```
std::wstring str = std::wstring_convert<std::codecvt_utf8<wchar_t>>().from_bytes("some string");
```

In the example they make use of the standard library method [`std::wstring_convert`](https://en.cppreference.com/w/cpp/locale/wstring_convert) from the `<locale>` header. Unfortunately this method has been deprecated in C++17, though as one note points out, it's been moved to an annex, where it will remain until a suitable replacement is built. As such it is still effectively part of the standard, and will remain so for now.

Related to the internals of the encoding conversions is the [`<codecvt>`](https://en.cppreference.com/w/cpp/header/codecvt) header file, which is part of the Localization library. It provides a few standard conversions between the utf8/utf16 encodings.

DanRStevens:
This closes #148.

I experimented with the standard C++ library methods for converting between UTF-8 and UTF-16 strings. It went quite well, so I decided to go with that. The change also helped remove Windows specific types from the function signatures, which allowed for the removal of the `<windows.h>` dependence.

### DIFF
--- a/srcStatic/LocalResource.h
+++ b/srcStatic/LocalResource.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <type_traits>
 
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -23,6 +23,15 @@ std::string ConvertWideToNarrow(std::wstring_view inputWideString)
 	return converter.to_bytes(std::wstring(inputWideString));
 }
 
+std::wstring ConvertNarrowToWide(std::string_view inputNarrowString)
+{
+	// Wide strings might be `wchar_t` or `char16_t`, so specify the type we want
+	// Create a UTF-8 <=> UTF-16 converter
+	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+	// Convert from narrow to wide
+	return converter.from_bytes(std::string(inputNarrowString));
+}
+
 std::size_t CopyStdStringIntoCharBuffer(const std::string& stringToCopy, char* buffer, std::size_t bufferSize)
 {
 	// Precheck valid pointer and non-zero buffer size to avoid wrap around or null termination problems
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -2,6 +2,7 @@
 #include <sstream>
 #include <algorithm>
 
+
 std::string WrapRawString(LPCSTR str)
 {
 	return std::string(str);
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -14,22 +14,22 @@ std::wstring WrapRawString(const wchar_t* str)
 	return std::wstring(str);
 }
 
-std::string ConvertWideToNarrow(const std::wstring& inputWideString)
+std::string ConvertWideToNarrow(const std::wstring& inputString)
 {
 	// Wide strings might be `wchar_t` or `char16_t`, so specify the type we want
 	// Create a UTF-8 <=> UTF-16 converter
 	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
-	// Convert from wide to narrow
-	return converter.to_bytes(inputWideString);
+	// Convert string
+	return converter.to_bytes(inputString);
 }
 
-std::wstring ConvertNarrowToWide(const std::string& inputNarrowString)
+std::wstring ConvertNarrowToWide(const std::string& inputString)
 {
 	// Wide strings might be `wchar_t` or `char16_t`, so specify the type we want
 	// Create a UTF-8 <=> UTF-16 converter
 	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
-	// Convert from narrow to wide
-	return converter.from_bytes(inputNarrowString);
+	// Convert string
+	return converter.from_bytes(inputString);
 }
 
 std::size_t CopyStdStringIntoCharBuffer(const std::string& stringToCopy, char* buffer, std::size_t bufferSize)
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -1,6 +1,6 @@
-#include "StringConversion.h"
 // Deprecation warning suppression must come before any standard library includes
 #define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
+#include "StringConversion.h"
 #include <sstream>
 #include <algorithm>
 #include <locale>
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -1,6 +1,8 @@
 #include "StringConversion.h"
 #include <sstream>
 #include <algorithm>
+#include <locale>
+#include <codecvt>
 
 
 std::string WrapRawString(const char* str)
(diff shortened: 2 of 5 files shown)


Brett208:
MSVC is giving the following warnings:


Warning	C4996	'std::wstring_convert<std::codecvt_utf8_utf16<wchar_t,1114111,0>,wchar_t,std::allocator<wchar_t>,std::allocator<char>>::from_bytes': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	op2extStatic	\op2ext\srcstatic\stringconversion.cpp	31

Warning	C4996	'std::codecvt_utf8_utf16<wchar_t,1114111,0>': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	op2extStatic	\op2ext\srcstatic\stringconversion.cpp	20	

Warning	C4996	'std::wstring_convert<std::codecvt_utf8_utf16<wchar_t,1114111,0>,wchar_t,std::allocator<wchar_t>,std::allocator<char>>': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	op2extStatic	\op2ext\srcstatic\stringconversion.cpp	20

Warning	C4996	'std::wstring_convert<std::codecvt_utf8_utf16<wchar_t,1114111,0>,wchar_t,std::allocator<wchar_t>,std::allocator<char>>::to_bytes': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	op2extStatic	\op2ext\srcstatic\stringconversion.cpp	25

Otherwise seems to compile and run okay. Tested by loading color mod 1 in debug compilation.

DanRStevens:
Ok, thanks for bringing up the big ugly warning that was generating. I suppose I shouldn't be surprised by the deprecation warning, considering the note about deprecation came up in the initial reading I did about the standard library features.

I opted to silence the warning. I figured that was consistent with my decision to use the feature despite knowing it had already been deprecated. As was stated elsewhere, a deprecated feature is still part of the standard, and although subject to potential removal from future standards, removal is not guaranteed, and is probably even unlikely until a suitable replacement is available.

We can still of course switch back to using [`WideCharToMultiByte`](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte) if we wish to avoid the deprecation. One point about the old code is it used `CP_ACP`, rather than `CP_UTF8`. The documentation provides some warning notes concerning that. If we go back to `WideCharToMultiByte`, we'll probably want to update the `CodePage` parameter.

PR #159, opened 2019-09-27
Cleanup whitespace constant

DanRStevens:
Replaced whitespace literals with named constant. Used a `std::string_view` to prevent expensive temporary object construction at point of calls for defaulted parameter uses.

### DIFF
--- a/srcStatic/StringConversion.cpp
+++ b/srcStatic/StringConversion.cpp
@@ -105,7 +105,7 @@ std::string TrimBack(const std::string& stringToTrim, std::string_view whitespac
 }
 
 
-std::vector<std::string> Split(std::string stringToSplit, char delimiter)
+std::vector<std::string> Split(const std::string& stringToSplit, char delimiter)
 {
 	std::vector<std::string> strings;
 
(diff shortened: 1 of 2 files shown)


srcStatic/StringConversion.cpp
```diff
@@ -70,7 +70,7 @@ std::string ToLower(std::string x) {
 
 
 // Trim whitespace from both ends of string, returning copy of substring
-std::string Trim(const std::string& stringToTrim, const std::string& whitespace)
+std::string Trim(const std::string& stringToTrim, std::string_view whitespace)
```
Brett208: I'm guessing we should start using string_view in a lot of other places and in OP2Utility instead of const std::string& all the time...

DanRStevens: Probably yes. In some cases it may not matter. There may be a few pitfalls with return values and object lifetime. One interesting benefit is it could increase the possibility of `constexpr` string functions, which could compute stuff at compile time. Certainly some nice to haves.

I'll open an Issue so we can track this and review the code later on for more `string_view` opportunities.

PR #160, opened 2019-09-27
Multiple console modules interface

DanRStevens:
This work supports Issue #100.

This makes the interface of the `ConsoleModuleLoader` class more closely match that of `ModuleLoader`. In particular, both now have an interface to support multiple modules.

Currently however, the `ConsoleModuleLoader` class internally still only supports a single module, and will raise an exception if you try to pass multiple modules.

Some calling code has been update to work as if multiple modules are already supported. Other places, notably `GetConsoleModDir_s` have been changed to assume they refer to console module at index 0. There is no workaround, other than to supply a new method and deprecate the old method.


srcDLL/Globals.cpp
```diff
@@ -7,5 +7,5 @@
 bool modulesRunning = false;
 
 VolList volList;
-ConsoleModuleLoader consoleModLoader(FindModuleDirectory());
+ConsoleModuleLoader consoleModLoader({FindModuleDirectory()});
```
Brett208: What does it mean to add the curly braces around FindModuleDirectory here?

Brett208:
I was thinking about the breaking change to the DLL interface. I'm fine with only passing the index 0 module with the existing function when it comes up in the future. 

Will do a more thorough checkout of the code in the near future. 

Thanks,
Brett

srcDLL/Globals.cpp
DanRStevens: The constructor now takes `std::vector` rather than `std::string`. The braces cause a vector to be constructed with that one element.

It's a bit of a temporary hack until we get command line parsing code that will return a vector of module names instead of just a single one.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -12,8 +12,19 @@
...
+// Console module names are the relative path from the game folder (no trailing slash)
+ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleNames)
 {
+	if (moduleNames.empty()) {
+		return; // No console modules to load
+	}
+
+	// For now just handle the first name
+	if (moduleNames.size() > 1) {
+		throw std::runtime_error("ConsoleModuleLoader currently only supports a single loaded module");
```
Brett208: This exception would be unhandled if it were thrown. It would be thrown during DLL initialization, so not really a way to catch it.

It should never be thrown because line 21 of ConsoleArgumentParser will abort when it detects multiple modules being loaded.

```C++
PostErrorMessage(__FILE__, __LINE__, "Error parsing command line arguments: " + std::string(e.what()));
```

This makes it seem a bit sloppy though. 

I understand this leaves the code in an intermediate state so I don't want to spend lots of time finding the perfect answer. If this used the PostErrorMessage and then returned without loading a module, it would keep from throwing the unhandled exception. @DanRStevens, what do you think?

DanRStevens: Well, yes, it is a bit sloppy. Though that is kind of by design. There is of course no intention to leave the code in that state. The check was added because I had chosen to defer work on the internals of the class until a later branch. The check is a temporary reminder that the guts of the class haven't yet been updated to support multiple modules. Only the interface has been updated so it could potentially support multiple modules.

More to the immediate point of the refactoring done in this branch, the constructor will only ever be passed a vector of 1 element. The caller is hardcoded to construct a vector of 1 element. That means the check will never be triggered in a production environment.

Ultimately the limitation on the constructor call does come from the `ConsoleArgumentParser` design. The function signature is set to return only a single value. Hence when that value is passed to the `ConsoleModuleLoader`, it is wrapped into a vector of 1 element, just to satisfy the interface change.

The plan is to update the internals before updating the command line argument processing.

If for some reason we do start passing multiple modules before the internals support it, that is a pretty serious logic flaw, and you might argue the program should crash (in a controlled manner).

Or maybe I'm just going about things backwards by updating the interface before updating the internals.

DanRStevens:
I would very much like to change to:
```
extern ConsoleModuleLoader consoleModuleLoader;
```

That has been bothering me as well.

srcStatic/ConsoleModuleLoader.cpp
Brett208: I can accept that this will be quickly modified again as an intermediate step.

I suppose terminating the program with an unhandled exception might be okay here (although I don't think I would prefer it), but it would still at least need to be logged why the program was terminated.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -47,16 +64,18 @@ std::size_t ConsoleModuleLoader::Count()
 	return moduleName != "" ? 1 : 0;
 }
 
+// Returns false if passed an empty string (Module name cannot be empty)
 bool ConsoleModuleLoader::IsModuleLoaded(std::string moduleName)
 {
-	if (Count() == 0) {
-		return false;
+	for (std::size_t i = 0; i < Count(); ++i) {
+		if (ToLowerInPlace(moduleName) == ToLower(GetModuleName(i))) {
```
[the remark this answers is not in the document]
Brett208: Looks like this optimization was made in the next branch, this comment can be ignored.

Issue #155, opened 2019-09-24
Remove GlobalDefines.h/.cpp


DanRStevens:
This file contains what are basically logging and reporting type methods. These methods should be moved to Log.h/.cpp instead, and the GlobalDefines.h/.cpp files should be removed.

During this cleanup, we should look at perhaps renaming some of the methods. Perhaps:
`OutputDebug` => `LogDebug`
`PostErrorMessage` => `LogError`

We should also take a look at `DisableModalDialogs` and see if we still need to keep this. Perhaps a change similar to what was done with `Log` could be used here.

Brett208:
The wording PostErrorMessage indicates a modal dialog appears. Are you suggesting, we just log the error and eliminate using the modal dialog with the name change?

I would definitely support changing from OutpostDebug to LogDebug. I'm not really a fan of providing Output to the IDE if we can just log. (There might be a few questionable calls to OutputDebug that should be removed from the code base, I think I removed them in the past).

DanRStevens:
Yes, there is some semantic change in meaning with the proposed rename.

I was thinking we could offer a `LoggerMessageBox`, which could be initially set to handle `LogError` output. The different logging methods could have different loggers set, to account for different levels of logging. It should also be fine if they are set to use a shared logger instance for multiple methods. The code is already using non-owning pointers.

The debug output could also have it's own logger object `LoggerDebug`.

We might also want a `Logger` object that distributes messages to multiple other loggers. That would allow logging to both a message box and to a log file, without having to tie that directly in to message box logging. Not too sure what to call it, maybe `LoggerDistributor`, or `LoggerTee` (though that name kind of implies 1 output to 2 outputs, rather than 1 to many).

The `LoggerMessageBox` and `LoggerDebug` classes would likely be default constructible. That means we could potentially create them on the fly quite easily within the logging code if no logger has been set when they are called. We may want to discuss that though. We probably also don't want to auto create if the user has specially set a `nullptr` for that logging level.

The `LoggerFile` object is not default constructible, due to the required `fileName` parameter. If the logging code is ever reused in another project, I'm not sure there would be a sensible default for that parameter, so I'd want to avoid automatic creation of `LoggerFile` within the logging code.

----

[...]

Issue #19, opened 2018-12-24
Automate DLL Unit Tests


DanRStevens:
Current testing for op2ext is through an ad hoc manual process. We should try to automate the testing, and have testing done on the CI build machines.

Testing of op2ext is a little more peculiar than testing static library code, or unit testing executable code:
Tester.exe -> libraryCode

In normal operation, the main executable for op2ext is Outpost2.exe, which we don't have source for, nor is it easy to modify, nor is it easy to automate. Outpost2.exe loads op2ext, though doesn't so much use op2ext, as have op2ext then load other 3rd party modules and provides an interface to them:
Outpost2.exe -> op2ext.dll <--> 3rdPartyModule.dll

I believe we can still use the normal testing process by building an executable, which loads the DLL, and then call functions from it as if it were a client loaded by op2ext:
op2extTester.exe -> op2ext.dll

We may need to account for any differences that op2ext assumes about it's runtime environment, such as patches made to Outpost2's address space. I see two ways we could go here. One is to modify op2ext to make fewer assumptions about its runtime environment, and thereby make it easier to test. This might be considered part of good design. Though some changes may be difficult or unsatisfying to make. How do we conditionally patch the exe, yet still have a robust and tested patching system that doesn't silently fail in real use? Another approach is to load Outpost2.exe into the address space of the test executable.

----

I have a few thoughts on loading Outpost2.exe into the address space of a test executable. First off, the test executable would be the main process host, not Outpost2.exe. As DLLs and EXEs have the same file format, I believe it would be possible to call [`LoadLibrary`](https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-loadlibrarya) on Outpost2.exe. I don't believe we need to actually run anything from Outpost2.exe, we just need it to be in memory. This CodeProject article may help: [Load EXE as DLL: Mission Possible](https://www.codeproject.com/Articles/1045674/Load-EXE-as-DLL-Mission-Possible).

We'll probably need to set a custom base address for the test executable, and move it somewhere other than the default of 0x400000. The image for Outpost2.exe is already set to be based at that address. As we don't have the source code for Outpost2.exe, we can't exactly recompile it, and when it was compiled, it would, by default, have it's relocations stripped. As is addressed in the article, DLLs are normally relocatable, with a relocation table, while EXEs are normally not relocatable, and have their relocation table stripped. This should be fine so long as we move the base address of the test executable so that it doesn't conflict.

[...]

PR #162, opened 2019-09-28
Remove GlobalDefines

DanRStevens:
This change supports Issue #155. There's still a bit of cleanup to do, though this seems like a good stopping point which takes care of most of the major points of the issue.

This cleanup may also help to support Issue #19, where popup boxes can be an impediment to automated testing.


srcStatic/Log.h
```diff
@@ -7,4 +7,19 @@ class Logger;
...
+void LogDebug(const std::string& message);
+
+// Logs an error message with the logger and then posts it to user in a modal dialog box.
+void PostErrorMessage(const std::string& errorMessage, const std::string& sourcePathFilename, long lineInSourceCode);
+
+
+// Allow defaulting of the __FILE__ and __LINE__ macros at the point of call
+// Macro name must be different from original function name
+// Using the same name will cause an error when the compiler sees the underlying function definition
+#define PostError(errorMessage) PostErrorMessage(errorMessage, __FILE__, __LINE__)
```
Brett208: I would be okay removing the idea of recording the file and code line number the error comes from. While it can be helpful, it isn't worth keeping around if it gets in the way of a quality log system.

srcStatic/Log.h
```diff
@@ -7,4 +7,19 @@ class Logger;
 
 
 void SetLogger(Logger* logger);
+void SetLoggerError(Logger* newLogger);
```
Brett208: Could we add a comment explaining what the difference between a Logger and LoggerError is. Based on reading what the issues we have on GitHub, I know what you are looking for, but just seeing this line of code I think would confuse me. The function names are not enough to distinguish to me.

Also, would recommend standardizing use of argument name to match between the two functions (whether the new prefix is kept or not).

DanRStevens: Ahh, thank you for noticing that. Seems I only updated it in the .cpp file.

srcStatic/Log.h
DanRStevens: I would like to keep the information. Maybe there's a way we could keep it that avoids irregularities in the API. Right now it creates irregularities in the naming of the core methods, and the number and types of parameters they accept. I guess I would be open to removing it as well. Probably a topic for a future branch.

DanRStevens:
You're right, we should try to reduce multiple pop-up errors to a single pop-up message box. I noticed this as well in the DLL test project when op2ext fails the initial patching of Outpost2.exe. I'll create an Issue to track this.

PR #164, opened 2019-09-28
Add DLL test project

### LINKED ISSUE #19: Automate DLL Unit Tests

DanRStevens:
This provides a starting base for Issue #19. There is still a bit more work to be done in porting test code over to this project. For now though, it serves as an example of how we might test the DLL.

Currently there are some error pop-up message boxes. These are an impediment to automation. They also pose a problem for runs on CircleCI, which does not have a GUI environment. Trying to create GUI objects in a non-GUI environment will cause the DLL unit test process to be aborted with an error. This results in test failures. For this reason, running of the DLL unit test project on CircleCI has been disabled.

Similarly, running the DLL test code on AppVeyor is expected to cause either similar problems, or a test hang, and eventual timeout. If we create a Visual Studio project file for the DLL unit test code, we'll need to ensure AppVeyor doesn't try to run it as part of the build and test process. Building the DLL test project should be fine, just not running it.

----

Edit: This closes #19.


Brett208:
Are the pop-up error messages being caused by our codebase (PostErrorMessage) or by the Linux operating system?

Let me know when you think it is mature enough and want a VS project file setup.

-Brett

DanRStevens:
Actually, we could probably add a project file now and merge it.

I was just looking at the AppVeyor config, and the test section runs the static library unit tests specifically, so it wouldn't try to auto run a new unit test project for the DLL. It builds everything, but only runs the static library unit tests.
```
build:
  project: op2ext.sln
test_script:
  - cd %APPVEYOR_BUILD_FOLDER%\test
  - '%APPVEYOR_BUILD_FOLDER%\%CONFIGURATION%\test.exe'
  - cd %APPVEYOR_BUILD_FOLDER%
```

DanRStevens:
Aside from the deprecated `GetCurrentModDir` function, which allocates a buffer than can never be safely deleted, all of the public methods exported by op2ext.dll are now tested.

Since the test environment used here does not try to load any modules, the methods that return information about loaded modules can only have limited testing. For now, more detailed module information testing is covered through manual testing with the TestModule project.

Pop-up message boxes that affect DLL unit testing have been removed. These unit tests can now be run on the continuous integration servers. This has already been setup on CircleCI.

The DLL unit test project still needs a Visual Studio project file. Testing on AppVeyor is currently blocked by the lack of a project file.

Brett208:
I'll look into creating this in the near future.

I haven't looked closely, but my first question will be if the project can be contained in the main solution or if it needs to be separate from all the other projects to allow appveyor to run and test it.

DanRStevens:
There should be no worries about AppVeyor. We can just create a new project within the same solution. There are no longer any modal pop-up messages to interfere with testing.

And also, AppVeyor won't even run the new project yet, since the AppVeyor config for testing explicitly runs only the static library unit tests. We would have to update it to also run the DLL unit tests.

Brett208:
MSVC project testDll produces the following warnings after debug compilation:

Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	60
Warning	C4309	'initializing': truncation of constant value	testDll	\op2ext\testdll\op2ext.test.cpp	10
Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	38
Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	39
Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	42
Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	59
Warning	C4068	unknown pragma	testDll	\op2ext\testdll\op2ext.test.cpp	63
Warning	C4309	'initializing': truncation of constant value	testDll	\op2ext\testdll\op2ext.test.cpp	73
Warning	C4309	'initializing': truncation of constant value	testDll	\op2ext\testdll\op2ext.test.cpp	168

testDll/op2ext.test.cpp
```diff
@@ -0,0 +1,188 @@
+#include "op2ext.h"
+#include <gtest/gtest.h>
+#include <string_view>
+#include <fstream>
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+
+
+TEST(op2ext, GetGameDir_s) {
+	constexpr char Nonce = 255;
```
Brett208: I'm a bit confused by what is meant by Nonce

testDll/op2ext.test.cpp
```diff
@@ -6,6 +6,10 @@
 #include <windows.h>
 
 
+// MSVC C4068 (unknown pragma) is issued on calls to #pragma GCC
+#pragma warning(disable: 4068)
```
DanRStevens: Ahh, we should maybe guard that pragma so it's only seen by Mingw, rather than disable MSVC warnings about it.

testDll/op2ext.test.cpp
DanRStevens: It's a fixed known value so you can detect writes or non-writes to a field. In this case, it's being used to check that a null terminator is not written when there is no valid buffer to write to. It is also  used to detect when the null terminator is written, as without initialization to a known value, you can't be certain if a value was an uninitialized null, or an explicitly written null.

PR #168, opened 2019-09-29
Add `memoryPatchingEnabled`

DanRStevens:
This reworks memory patching enabling/disabling to be more automated. Instead of explicit calls to disable patching, it relies on the already present setup calls to enable it. This allows memory patching to be enabled exactly when it is needed.

Edit: Tagging Issue #19 as related.

### DIFF
--- a/srcStatic/OP2Memory.cpp
+++ b/srcStatic/OP2Memory.cpp
@@ -35,6 +35,7 @@ void SetLoadOffset()
 
 	if (op2ModuleBase == nullptr) {
 		PostError("Could not find Outpost2.exe module base address.");
+		return;
 	}
 
 	loadOffset = reinterpret_cast<std::size_t>(op2ModuleBase) - ExpectedOutpost2Addr;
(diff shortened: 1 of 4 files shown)


srcStatic/OP2Memory.cpp
```diff
@@ -6,7 +6,7 @@
 #include <iomanip>
 
 
-bool memoryCommandsDisabled;
+bool memoryPatchingEnabled = false;
```
Brett208: I like changing this variable from Disabled to Enabled. Makes the code easier to read.

srcStatic/OP2Memory.cpp
```diff
@@ -19,32 +19,26 @@ std::string AddrToHexString(std::size_t addr)
...
-
 	void* op2ModuleBase = GetModuleHandle(TEXT("Outpost2.exe"));
 
 	if (op2ModuleBase == nullptr) {
 		PostError("Could not find Outpost2.exe module base address.");
+		return;
 	}
 
+	// Enable memory patching for Outpost2.exe, and set relocation offset
+	memoryPatchingEnabled = true;
```
Brett208: This is much better placement than how I had it originally designed.

PR #169, opened 2019-09-29
Move `AddrToHexString` and `GetDateTime`

DanRStevens:
The `AddrToHexString` and `GetDateTime` methods are fairly general in their use. This moves them to a more publicly visible place.

Unit tests were added for these methods. The `GetDateTime` method was fixed so that it actually works on Linux with Mingw. Previously log messages had empty strings for the timestamps.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -9,7 +9,6 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <string>
-#include <sstream>
 
 
 void LocateVolFiles(const std::string& relativeDirectory = "");
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -74,9 +74,7 @@ int TApp::Init()
 	bool success = Op2UnprotectMemory(destinationBaseAddress, 0x00587000 - 0x00585000);
 
 	if (!success) {
-		std::ostringstream stringStream;
-		stringStream << "Error unprotecting memory at: 0x" << std::hex << destinationBaseAddress << ".";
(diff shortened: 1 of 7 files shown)


srcStatic/StringConversion.h
```diff
@@ -52,3 +52,10 @@ std::vector<std::string> SplitAndTrim(std::string stringToSplit, char delimiter,
 	}
 	return items;
 }
+
+// Convert hex address value to string
+std::string AddrToHexString(std::size_t addr);
```
Brett208: I wouldn't have thought to put AddrToHexString and GetDateTime in the StringConversion file. However, I think it works fine. 

I like that they are usable by the rest of the code base now.

srcDLL/DllMain.cpp
```diff
@@ -74,9 +73,7 @@ int TApp::Init()
 	bool success = Op2UnprotectMemory(destinationBaseAddress, 0x00587000 - 0x00585000);
 
 	if (!success) {
-		std::ostringstream stringStream;
-		stringStream << "Error unprotecting memory at: 0x" << std::hex << destinationBaseAddress << ".";
-		PostError(stringStream.str());
+		PostError("Error unprotecting memory at: 0x" + AddrToHexString(destinationBaseAddress));
```
Brett208: Happy to see this error message formatting simplified by moving the formatting code out of OP2Memory.

srcStatic/StringConversion.h
DanRStevens: Yes, I've been piling a lot of string methods into there. At some point maybe we should either split it, or rename it to better reflect what it contains.

PR #171, opened 2019-09-30
Build out multiple console modules support

DanRStevens:
This extends the `ConsoleModuleLoader` class to  add support for loading multiple console modules.

Not covered are updates to the command line argument processing, which would be needed to pass multiple module names to the `ConsoleModuleLoader`'s constructor.

This work supports Issue #100. This is the internal changes that correspond to the interfaces changes in PR #160.

### DIFF
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -97,12 +97,6 @@ void ConsoleModuleLoader::LoadModules()
 
 	// Load all module DLLs
 	for (auto& module : modules) {
-		std::error_code errorCode;
-		if (!fs::is_directory(module.directory, errorCode)) {
-			PostError("Unable to access the provided module directory. " + errorCode.message());
-			continue;
-		}
-
 		LoadModuleDll(module);
 	}
 }
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -23,6 +23,11 @@ ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleN
 		return; // No console modules to load
 	}
 
+	if (std::any_of(moduleNames.begin(), moduleNames.end(), [](const std::string& moduleName){ return moduleName.empty(); })) {
+		PostError("All console module names must be non-empty.");
+		return;
+	}
+
 	// For now just handle the first name
 	if (moduleNames.size() > 1) {
 		throw std::runtime_error("ConsoleModuleLoader currently only supports a single loaded module");
--- a/srcStatic/ConsoleModuleLoader.cpp
+++ b/srcStatic/ConsoleModuleLoader.cpp
@@ -39,7 +39,7 @@ ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleN
 	}
 
 	// Store module details
-	module = {nullptr, moduleName};
+	modules.push_back({nullptr, moduleName});
 	// Set private static instance by reference
 	ModuleDirectory() = moduleDirectory;
 }
@@ -55,15 +55,15 @@ std::string ConsoleModuleLoader::GetModuleDirectory(std::size_t index)
 
 std::string ConsoleModuleLoader::GetModuleName(std::size_t index)
 {
-	if (index >= Count()) {
+	if (index >= modules.size()) {
 		throw std::runtime_error("Invalid console module index: " + std::to_string(index));
 	}
-	return module.name;
+	return modules[index].name;
 }
 
 std::size_t ConsoleModuleLoader::Count()
 {
-	return module.name != "" ? 1 : 0;
+	return modules.size();
 }
 
 // Returns false if passed an empty string (Module name cannot be empty)
@@ -94,7 +94,11 @@ void ConsoleModuleLoader::LoadModules()
 	}
 
 	HookFileSearchPath();
-	LoadModuleDll(module);
+
+	// Load all module DLLs
+	for (auto& module : modules) {
+		LoadModuleDll(module);
+	}
 }
 
 void ConsoleModuleLoader::LoadModuleDll(Module& module)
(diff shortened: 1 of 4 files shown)


srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -18,57 +18,61 @@ ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleN
...
-		throw std::runtime_error("ConsoleModuleLoader currently only supports a single loaded module");
+	// Temporary check. This will eventually become an error.
+	if (moduleNames.size() == 1 && moduleNames[0].empty()) {
+		return; // No console modules to load
 	}
-	auto moduleRelativeDirectory = moduleNames[0];
 
-	if (moduleRelativeDirectory.empty()) {
-		return; // No Console Module Loaded
+	if (std::any_of(moduleNames.begin(), moduleNames.end(), [](const std::string& moduleName){ return moduleName.empty(); })) {
```
Brett208: This is a good, concise check to add.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -18,57 +18,61 @@ ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleN
...
+		PostError("All console module names must be non-empty.");
+		return;
 	}
 
-	auto moduleDirectory = fs::path(GetGameDirectory()).append(moduleRelativeDirectory).string();
-	moduleName = moduleRelativeDirectory;
+	for (auto& moduleName : moduleNames) {
```
Brett208: I would typically mark this const auto& to indicate no changes to container contents being looped through.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -18,57 +18,61 @@ ConsoleModuleLoader::ConsoleModuleLoader(const std::vector<std::string>& moduleN
...
 
-	std::error_code errorCode;
-	if (!fs::is_directory(moduleDirectory, errorCode)) {
-		PostError("Unable to access the provided module directory. " + errorCode.message());
-		moduleDirectory = "";
-		moduleName = "";
-	}
+		std::error_code errorCode;
+		if (!fs::is_directory(moduleDirectory, errorCode)) {
+			PostError("Unable to access the provided module directory. " + errorCode.message());
```
Brett208: It would be nice to provide the module name as part of the error message here now that multiple modules could produce the error.

srcStatic/ConsoleModuleLoader.cpp
```diff
@@ -175,35 +179,39 @@ bool ConsoleModuleLoader::ResManager::GetFilePath(const char* resourceName, /* [
...
 void ConsoleModuleLoader::RunModules()
 {
-	// Startup a module by calling its run func
-	if (modDllHandle)
-	{
-		// Call its mod_run func
-		FARPROC runFunc = GetProcAddress(modDllHandle, "mod_run");
-		if (runFunc) {
-			runFunc();
+	for (auto& module : modules) {
```
Brett208: I think this could be marked const auto& as well.

test/ConsolueModuleLoader.test.cpp
```diff
@@ -66,3 +66,34 @@ TEST(ConsoleModuleLoader, ModuleWithEmptyDLL)
...
+	EXPECT_TRUE(consoleModuleLoader.IsModuleLoaded(moduleNames[1]));
+
+	EXPECT_FALSE(consoleModuleLoader.IsModuleLoaded(""));
+
+	EXPECT_NO_THROW(consoleModuleLoader.LoadModules());
+	EXPECT_NO_THROW(consoleModuleLoader.RunModules());
+	EXPECT_NO_THROW(consoleModuleLoader.UnloadModules());
+
+	// // Cleanup test module directories
+	// for (const auto& moduleName : moduleNames) {
```
Brett208: I think I would prefer this be un-commented and guarded by a preprocessor directive to prevent it running on mingw but still let it cleanup on MSVC builds.

DanRStevens: I'm hesitant to jump straight to a compiler/platform detect solution. I'd like to experiment a bit more to see if I can get something working across all compilers. Plus, it would be nice to cleanup all of the test directory creation.

DanRStevens:
Ok, I'll check into it.

There was a change to the resource loading in this branch in commit https://github.com/OutpostUniverse/op2ext/pull/171/commits/b14d71960a4ae573c4aea7e87ce6241103283c94. The change mostly just wrapped existing code in a loop. I didn't see anything obviously wrong on quick inspection. Not sure where else the problem could have come from though.

----

For the `PostError` silencing, that likely stems from PR #141, and in particular commit https://github.com/OutpostUniverse/op2ext/pull/141/commits/1e81f58244ba0abb4f45bce876b04a60e2b85d48. I had noticed the silencing recently. We should fix it in another branch. I'll open an issue to track it. I think the fix should be fairly easy.

DanRStevens:
Ahh, the module directories do not include a trailing slash:
```cpp
for (const auto& moduleDirectory : moduleDirectories) {
	// Search for resource in module folder
	const auto path = moduleDirectory + resourceName;
	if (INVALID_FILE_ATTRIBUTES != GetFileAttributesA(path.c_str())) {
		if (0 == CopyStdStringIntoCharBuffer(path, filePath, MAX_PATH)) {
			return true; // Resource found
		} else {
			Log("MAX_PATH exceeded while trying to return path to resource: " + std::string(resourceName));
		}
	}
}
```

It uses `std::string` here, so it's simple pasting of:
```cpp
const auto path = moduleDirectory + resourceName;
```

Without a trailing slash on `moduleDirectory` the result is malformed.

DanRStevens:
I tried a quick fix here, but it wasn't working, so I rolled back the changes.

I've verified the bug is present on master. Perhaps the resource loading problem should be fixed in a separate branch.

DanRStevens:
My fault for not testing better. This seems to be one of the areas hit by bad code gen with Mingw, so I probably didn't test changes as well as I should have.

I opened Issue #175 to track the problem.

Issue #139, opened 2019-09-21
Allow loading external modules in the ini file via a full section


Brett208:
Currently external modules are loaded like:

[Game]
```
ExternalModules = "NetFix, NetHelper"
```

I wold prefer loading ExternalModules:
```
[ExternalModules]
NetFix = yes
NetHelper = no
```

However, I haven't researched Ini file access enough to know if this is supportable by the winapi. We can get away with this sort of list for BuiltInModules because we know all their names ahead of time. For an external module, we don't know the module's name before compilation.

DanRStevens:
I don't believe the Windows API would have support for using INI files that way. You could certainly do it if you wrote your own INI file parser though.

Of course, that does seem a little against the spirit of how INI files were meant to be used. They were about key/value lookup, for well known configuration settings. But then, I suppose we're already abusing the format with `ExternalModules` by having values that refer to section names.

Brett208:
At this time, I'm not looking to write a custom parser. I guess we can leave this issue on the backburner in case it ever gets realized as I think it would provide better presentation to the user.

DanRStevens:
I agree. It does look nicer, and makes it easier for users to turn modules on/off.

DanRStevens:
I just remembered something from years ago. We can do this without a custom parser. The [`GetPrivateProfileString`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilestring) function supports getting all section names, or all key names within a section.

> `lpAppName`
> 
> The name of the section containing the key name. If this parameter is NULL, the GetPrivateProfileString function copies all section names in the file to the supplied buffer.

> `lpKeyName`
> 
> The name of the key whose associated string is to be retrieved. If this parameter is NULL, all key names in the section specified by the lpAppName parameter are copied to the buffer specified by the lpReturnedString parameter.

That was the reason why there buffer size might need to be 2 larger than the string size. Because under certain conditions it can return an array of strings, both the strings and the array may need null termination.

This is similar to how passing `nullptr` in to `WritePrivateProfileString` can be used to clear a key or a section.

DanRStevens:
[...]

Issue #76, opened 2019-04-23
Add tests for IniModules


Brett208:
Currently, cannot test past having an empty/missing .ini file.

I think there are two choices here:
1. Dynamically try to write/update a test.ini file to pull values from during testing
2. Extract parsing the .ini file out of IniModuleLoader using Dependency Injection. Instead pass the results of parsing the .ini file into IniModuleLoader, allowing dummy settings to be passed in without actually reading a .ini file.

Dynamically re-writing the .ini file increases the amount of code tested as it also makes sure we properly read the ini file. There are some technical hurdles in that I don't know how easy it will be to write/rewrite the .ini file during unit tests. I also don't know how the appveyor automated build would respond to this.

Using dependency injection seems a little easier to implement from a technical perspective. I also like that it separates the code base into smaller parts.

DanRStevens:
3. Pass an IniFile object to the IniModuleLoader constructor. When testing, an IniFileMock object can be passed instead that returns a set of pre-programmed responses. This would require defining an IniFile interface, which passed objects must adhere to, and reworking internal ini file access to go through the interface. The IniFile interface would need to support functionality similar to `GetOP2PrivateProfileString`, and optionally could be extended further to restrict access to a specific IniSection, with similar functionality but fewer parameters.

DanRStevens:
Not really related to #131, but the Google Mock technique presented there may apply here.

Brett208:
Since the ini module loader no longer exists, I renamed this issue. The heart of being able to test an ini module still exists though.

Brett208:
Thanks to @DanRStevens' work with creating an IniFile abstraction and the improved ModuleLoader over IniModule, we are closer to testing. The next problem is not having a DLL to test against.

Can we place an empty DLL in the test directory to use in testing? Not sure if the mingw compiler can read a DLL formed via MSVC compilation. 

Another option would be to strip out portions of TestModule that are now tested using TestDll. Then we could compile TestModule as a DLL before running the test code and use it for IniModule testing.

I'm not sure if GoogleMock can be applied here. Frankly I'm not knowledgeable about mocking at all.

DanRStevens:
Mingw can compile DLLs, so we can produce test modules on Linux from source code. The makefile is configured so it can build the TestModule project. I've used it a few times from Linux to do some manual testing.

[...]

PR #172, opened 2019-10-01
Add `IniFile` class

DanRStevens:
This wraps up INI file use into a new `IniFile` class.

This sets a foundation to build from that could help with Issues #76 and #139.

### DIFF
--- a/srcStatic/IniFile.h
+++ b/srcStatic/IniFile.h
@@ -19,6 +19,8 @@ public:
 	void ClearSection(const std::string& sectionName);
 	void SetValue(const std::string& sectionName, const std::string& keyName, const std::string& value);
 
+	static std::string GetValue(const std::string& fileName, const std::string& sectionName, const std::string& keyName);
+
 private:
 	const std::string fileName;
 };
--- a/srcStatic/IniFile.h
+++ b/srcStatic/IniFile.h
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <string>
+
+
+class IniSection;
+
+
+// Represents an entire .ini file
+class IniFile {
+public:
+
+	IniFile(std::string fileName);
+
+	const std::string& FileName() const;
+	std::string GetValue(const std::string& sectionName, const std::string& keyName) const;
+	IniSection operator[](std::string sectionName) const;
+
+private:
+	const std::string fileName;
+};
+
+
+// Represents a single section within a .ini file
+class IniSection {
+public:
+	IniSection(std::string fileName, std::string sectionName);
+
+	const std::string& FileName() const;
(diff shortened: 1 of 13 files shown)


srcStatic/IniFile.h
```diff
@@ -0,0 +1,50 @@
...
+
+
+class IniSection;
+
+
+// Represents an entire .ini file
+class IniFile {
+public:
+
+	IniFile(std::string fileName);
```
Brett208: I would have preferred to see filename for the various argument and variable instance names (treating filename as one word instead of two words, file name).

srcStatic/IniFile.h
```diff
@@ -0,0 +1,50 @@
...
+
+	const std::string& FileName() const;
```
Brett208: I think this makes returning the filename dependent on not deconstructing the class since the saved instance of the string referenced is contained within the class?

This is probably an okay assumption to make, and I assume it provides a bit of optimization by not having to copy the filename when referenced. I'm not sure the 'risk' is worth the reward here considering the filename will likely not be referenced repeatedly in a performance hampering way. Although my thoughts on the matter I think are more of an opinion than anything else.

Same would apply to a couple of GET functions from the IniSection class.

srcStatic/IniFile.h
DanRStevens: I thought about that. I figured you might mention it. Though after writing "keyName" and "sectionName" so many times, it just felt wrong to write anything other than "fileName". In this context, I am trying to distinguish between different types of names, one of which happens to be of files. As such, I would consider this a non-compound form of "file name".

srcStatic/IniFile.h
DanRStevens: Yes, you're right. It ties the lifetime of the result value to the lifetime of the object. Yes, that can open up the possibility of a dangling reference, which the compiler is not likely to warn about.

The likelihood of code being written that triggers that behaviour seems small, and is easy enough to work around. I figured that solution was safe enough here since the uses are likely for immediate consumption in logging or display purposes. By returning a reference it can be done more efficiently, and by making it `const`, the internals of the class are still protected.

If the value needs to have its lifetime extended, it can always be copied to a new string when capturing the value. Example:
```cpp
std::string name = iniFile.FileName(); // Copy to new std::string
```

That way the calling code can decide if it wants efficiency, or generality. I figured it went along with the C/C++ ethos of not paying for things you don't need.

Also, this conversation very much applies to how we want to approach using `std::string_view` in the future. It's the exact same concept if you return a non-owning read-only `std::string_view` from a function. Although you need to beware, I think in some cases it's worth it. Subjective argument though of course.

PR #188, opened 2019-10-06
Add IniFile GetSectionNames and GetKeyNames

DanRStevens:
This provides support functionality for Issue #139.

### DIFF
--- a/srcStatic/IniFile.cpp
+++ b/srcStatic/IniFile.cpp
@@ -46,30 +46,18 @@ void IniFile::SetValue(const std::string& sectionName, const std::string& keyNam
 // This can be used for convenient one-off use
 // This may be less efficienct for repeated use to fetch multiple values
 std::string IniFile::GetValue(const std::string& fileName, const std::string& sectionName, const std::string& keyName) {
-	const std::size_t bufferInterval = 1024;
-	auto currentBufferSize = bufferInterval;
-	std::string profileString;
-	DWORD returnSize;
+	// Allocate buffer space
+	// The Win API GetPrivateProfileString has a maximum limit of 2^16 characters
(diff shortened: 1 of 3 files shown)


srcStatic/IniFile.cpp
```diff
@@ -87,6 +84,39 @@ void IniFile::SetValue(const std::string& fileName, const std::string& sectionNa
...
 }
 
+// Delegate to Windows API to read INI file
+// Wrap result in std::string
+std::string IniFile::GetIniString(const char* fileName, const char* sectionName, const char* keyName) {
+	// Allocate buffer space
+	// The Win API GetPrivateProfileString has a maximum limit of 2^16 characters
+	// Trying to retrieve a longer string from the Win API will fail in unusual ways
+	std::string resultString;
+	resultString.resize(65536);
```
Brett208: I had chosen 1024 as the initial buffer size because it seemed large enough to catch most any ini setting. If you think casting to 65536 and then truncating later is more efficient and easier to read, I will defer to your judgement. I'm not in the mood to profile between the two choices as long as it works and there is no noticeable lag.

DanRStevens: Indeed, 1024 is pretty generous, and not likely to be a performance burden either.

As a bit of an aside, I was curious about the runtime too. I noticed Google Test results show the runtime for various test cases. I looked at it briefly after making this change, and noticed the affected tests had about the same or lower runtimes. I'm not certain the handful of runs I did was statistically significant though, and the test times are likely dominated by going to disk, rather than anything to do with looping or not looping. Could just have been a few lucky runs, or disk caching effects. I suspect a real speed test would show no meaningful performance difference.

At any rate, the allocation won't actually initialize the full buffer, just allocate space for it. Allocation time should be largely the same, regardless of how much is being allocated.

Issue #167, opened 2019-09-29
Report errors in a single pop-up message box


DanRStevens:
There are a few places in the code with pop-up message box errors at multiple levels. This can cause multiple pop-up message boxes to appear for a single error. We should reduce the reporting to a single pop-up message box for what is conceptually a single error.

One way to do this, is to replace the lower level error reporting with exceptions, and then collate the error messages at the higher level to report them. This can be done by effectively wrapping error messages, so when a higher level error is thrown, the text includes the `exception.what()` text from the lower level error.

More long term, it would be nice to move away from modal pop-up boxes that require the user to click close. Such model boxes can be annoying to users, and are also an impediment to automated testing.

Brett208:
I agree with consolidating modal dialogs so multiple do not appear for a single issue. 

I'm not sure I would support removing the modal dialogs unless there is a different way to directly and immediately let the user know they are experiencing a serious problem. For example with the module NetHelper, everything is behind the scenes. The user would not know there was a problem with the module unless they search through the log files.

DanRStevens:
Immediate visual feedback is fine. Just as long as it's non-modal. There should be no required user interaction to dismiss the error, or to continue. Examples would be asynchronous system tray popup notifications, which auto dismiss themselves after some timeout, or an overlay log display at the main game menu.

PR #190, opened 2019-10-06
Create console module class

### LINKED ISSUE #117: Unify ModuleLoader
I would like to consider unifying the 2 existing game module types and adding a third module type using inheritance. A base abstract class called GameModule and an inheritance tree of modules that are all handled by the same ModuleManager after being loaded.

```C++
#include <string>

// Abstract class allowing memory modification of Outpost 2 to alter gameplay or fix bugs
class GameModule
{
public:
    GameModule(const std::string name) : name(name);
    virtual ~GameModule() 
    inline std::string Name() { return name; } const;

    virtual void Initialize(); //mod_init(), InitMod
    // Only available for Console modules (other modules would just call an empty function)
    virtual void Run() = 0; //mod_run()
    virtual void Destroy() = 0; //mod_destroy(), DestroyMod

private:
    const std::string name;
}
```

```C++
#include <windows.h>
#include <string>

// Game Module identified by command line switch. Use for modules that should only occasionally be loaded from the command line.
class ConsoleModule : GameModule
{
public:
    ConsoleModule(const std::string& name, const std::string& moduleDirectory) : GameModule(name), moduleDirectory(moduleDirectory);

    std::string GetModuleDirectory() const;

    void Load() override;
    void Initialize() override;
    void Destroy() override;

private:
    const std::string moduleDirectory;
    HINSTANCE dllHandle;
}
```

```C++
// Game module identified in INI file with logic contained in a separate dll. Use for large modules  designed for semi-permanent use.
class IniModule : GameModule
{
}

// Internal modules. Use for small patches where creating a full new project and dll are not practical.

class IPDropDown : GameModule
{
}

class GarageSaveCrash : GameModule
{
}

class BuildingGroupHang : GameModule
{
}

// and so on...
```

For cleanup, ini modules and console modules use the function Destroy. I would prefer the word Unload over Destroy, but maybe best to stick with Destroy since the EXPORT functions use it.

My biggest concern is dealing with memory modification of functions from within the classes. When attempting to patch Outpost2 from a class called GarageSaveCrash, if the function is defined within the class, it will include the hidden argument detailing which class the function is a part of. I don't know if we can easily overcome this.

Once solution could be to define replacement functions in the class's cpp file but outside of the class. This seems maybe a bit contrived/awkward though.

If this has merit, it would need more fleshing out. I'd be willing to do a lot of leg work although someone else would have to help me with the memory hooks.

Thanks,
Brett

Brett208:
Closes #117 

Contains breaking changes to the external API in op2ext.h by removing the functions to return the console module and ini module count individually. I think this is acceptable because:

1. An external module user shouldn't need to worry how other modules are loaded, only that it is loaded.
2. We added this function in the last release and I doubt anyone has programme against it (except maybe Crow!).
3. I think it was a mistake I made the functions separate in the first place.

I tested this using the TestModule and cm1. I would like to test resource loading a little more before merging, although it looks like vol loading is working.

-Brett


srcDLL/op2ext.cpp
```diff
@@ -115,23 +115,13 @@ OP2EXT_API void Log(const char* message)
 
 
 OP2EXT_API bool IsModuleLoaded(const char* moduleName)
-{
-	return IsConsoleModuleLoaded(moduleName) || IsIniModuleLoaded(moduleName);
-}
-
-OP2EXT_API bool IsConsoleModuleLoaded(const char* moduleName)
```
Brett208: A breaking change to the API by removing this function and IsIniModuleLoaded. User would now be forced to call IsModuleLoaded and would not know if it was loaded as an ini or console module.

srcStatic/GameModules/ConsoleModule.h
```diff
@@ -0,0 +1,28 @@
...
+	inline std::string Directory() { return moduleDirectory; }
+
+private:
+	typedef void(*LoadModuleFunction)();
+	typedef bool(*UnloadModuleFunction)();
+	typedef void(*RunModuleFunction)();
+
+	std::string moduleDirectory;
+
+	LoadModuleFunction loadModuleFunction = nullptr;
```
Brett208: I wanted to combine these external function hooks with their ini module equivalents and place them in the abstract class DllModule. However, LoadModuleFunction includes an argument in the ini version. The other two, Unload and Run could be combined, but I didn't know if that made since to leave Load orphaned.

I set these = nullptr because the module may not have an associated dll. Otherwise, I'm worried C++ will leave garbage allocated instead of a null value, causing problems when the module's load, run, and unload functions are called.

DanRStevens:
Normally I'm hesitant to change already published interfaces, though I think I'm in agreement on this one. Your reasons are sound, and we probably shouldn't be too hesitant to change new features that perhaps haven't stabilized yet.

We be good to involve Crow in the discussion on this.

Looks like some test failures on CircleCI due to uncaught exceptions.

I'll review shortly.

srcStatic/GameModules/ConsoleModule.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "../Log.h"
+#include <windows.h>
+#include <stdexcept>
+
+ConsoleModule::ConsoleModule(const std::string& moduleName) : DllModule(moduleName)
+{
+	moduleDirectory = (fs::path(GetGameDirectory()) / moduleName).string() + "\\";
+
+	std::error_code errorCode;
+	if (!fs::is_directory(moduleDirectory, errorCode)) {
```
DanRStevens: This is likely to create problems for Mingw. The `is_directory` method will return false for any paths with a trailing slash. It maybe be easier to append the slash after checking `is_directory`.

Alternatively, perhaps we should create an `IsDirectory` method in FileSystemHelper, which contains logic to work around Mingw bugs.

srcStatic/GameModules/ConsoleModule.h
DanRStevens: Load differs by a parameter, Unload differs by a return type. The return type isn't really used though, so we could sort of combine things there. Might make sense to log if the return value of Unload indicated an error.

To truly clean this up and unify things, we'd probably need to create a new specification that wasn't backwards compatible.

The `nullptr` initialization looks like a good idea.

srcStatic/ModuleLoader.cpp
```diff
@@ -74,6 +112,21 @@ std::string ModuleLoader::GetModuleName(std::size_t index)
 	return modules[index]->Name();
 }
 
+std::string ModuleLoader::GetModuleDirectory(std::size_t index)
+{
+	if (index >= modules.size()) {
+		throw std::out_of_range("GetModuleDirectory: Invalid console module index: " + std::to_string(index));
```
DanRStevens: We might want to drop "console" from this error message.

srcStatic/ModuleLoader.cpp
```diff
@@ -74,6 +112,21 @@ std::string ModuleLoader::GetModuleName(std::size_t index)
...
+	}
+
+	ConsoleModule* consoleModule = dynamic_cast<ConsoleModule*>(modules[index].get());
+	if (consoleModule != nullptr) {
+		return consoleModule->Directory();
```
DanRStevens: This is correct, though I might prefer is the `dynamic_cast` was not required. We could get rid of it if the base class guaranteed to provide a `Directory()` method. The base class could have a default implementation that returns the empty string.

srcStatic/ModuleLoader.cpp
```diff
@@ -105,8 +158,19 @@ void ModuleLoader::RegisterModule(std::unique_ptr<GameModule> newGameModule)
...
+	RegisterConsoleModules();
 	RegisterIniModules();
 
+	// Abort early to avoid hooking file search path if not needed
+	if (modules.empty()) {
+		return;
+	}
+
+	// Setup loading of additional resources from module folders
+	ResourceSearchPath::Activate();
```
DanRStevens: Looks like the `ResourceSearchPath` is now being activated if any modules are loaded, not just if any console modules are loaded. This may pose a problem on Mingw, as the current hook tries to chain to the original Outpost2.exe method when the augmented path doesn't find a file. That chaining to the original method was affected by a Mingw code gen bug. This will need some additional testing.

DanRStevens: Now that unit tests are passing, I gave this a try. The game does indeed crash now on Linux if any modules are loaded. Previously it only crashed if `/loadmod` was used. Now it crashes all the time since there are BuiltIn modules or IniModules that get loaded and cause `ResourceSearchPath::Activate()` to be called.

Maybe that call can be moved into the Console module loading code for now, and made conditional on console modules being present.

----

Side note: Initially, I was confused by this not crashing as I expected. This was until I realized I was testing on my old stock INI file, but recent changes have renamed certain keys used for module loading. I had to create an "ExternalModules" key to get INI file loading working again before I could see this crash. This is a reminder that although INI keys might be renamed in this project, the game is packaged separately, and corresponding updates would need to be made there.

srcStatic/ModuleLoader.cpp
```diff
@@ -26,6 +29,41 @@ void ModuleLoader::RegisterBuiltInModules()
...
+
+	if (std::any_of(consoleModuleNames.begin(), consoleModuleNames.end(), [](const std::string& consoleModuleName) { return consoleModuleName.empty(); })) {
+		PostError("All console module names must be non-empty.");
+		return;
+	}
+
+	std::vector<std::string> moduleDirectories;
+	moduleDirectories.reserve(modules.size());
+
+	for (const auto& moduleName : consoleModuleNames)
```
DanRStevens: To be consistent with the old code, the `reserve` should perhaps be using `consoleModuleNames.size()`.

There may be cause to update this to include directories for other module types so this isn't necessarily bad. Though I also had some thoughts to allowing multiple `ResourceSearchPath` objects, which could be loaded and unloaded independently. That would allow for separate instances to manage search paths for console modules versus other sources. Hence we may want to keep this tied to the number of console modules anyway.

srcStatic/ModuleLoader.cpp
DanRStevens: Update: If you merge PR #196 first, it should mitigate this issue.

srcDLL/DllMain.cpp
```diff
@@ -85,16 +83,15 @@ int TApp::Init()
...
 	moduleLoader->LoadModules();
 
 	// Find VOL files from additional folders
-	for (std::size_t i = 0; i < consoleModuleLoader->Count(); ++i) {
-		// ConsoleModule name matches relative folder from game exeucutable folder
-		LocateVolFiles(consoleModuleLoader->GetModuleName(i));
+	for (std::size_t i = 0; i < moduleLoader->Count(); ++i) {
+		std::string moduleDirectory = moduleLoader->GetModuleDirectory(i);
+		if (moduleDirectory != std::string()) {
+			LocateVolFiles();
```
DanRStevens: Hmm, this call lost the parameter for the console module folder. The old corresponding line was:
```cpp
LocateVolFiles(consoleModuleLoader->GetModuleName(i));
```

Although tempting, it is not possible to use `moduleLoader->GetModuleDirectory(i)` here, since that returns an absolute path, whereas the `LocateVolFiles` function only accepts relative paths.

For now, I think this needs to be updated to:
```cpp
LocateVolFiles(moduleLoader->GetModuleName(i));
```

DanRStevens:
Ok, I've resolved all the issues I spotted here in another branch. You can merge PR #197 into this branch if you agree with all of it, or cherry-pick changes if you just want some of it.

DanRStevens:
Oh, and minor point, the `DllModule` class is abstract. It doesn't not implement all pure virtual methods of its base class `GameModule`. As such, it is more like `GameModule` than the concrete derived classes. It should probably be located in the repository in the same place as `GameModule`, rather than in the subfolder with the derived classes.

Not sure if we want to revisit that organization.

Brett208:
I'd prefer to leave DllModule in the GameModules subdirectory. 

The fixes branch has been merged in and now I believe this branch now closes #167.

-Brett

Issue #194, opened 2019-10-07
Extract TAppEvent hook code


DanRStevens:
Currently the TApp event hooking code is spread about DllMain.cpp. This code should be extracted to its own file, perhaps TAppEvent.h/.cpp.

Currently 3 events are hooked:
- TApp.Init
- TApp.ShutDown
- ShellLoaded (somewhat badly named as `NewLoadLibraryA`)

Currently patching is done in multiple places. The 3 events are hooked using two calls to `Op2RelinkCall`, and one call to `Op2MemSetDword`. The patching should be collected into a single method to install the event hooks. This method should be called from `DllMain`, rather than the more cryptic and low level `Op2RelinkCall` method.

Callbacks could be handled with lambdas or similar. In particular, the `NewLoadLibraryA` hook method is perhaps too low level to be placing actual event handling code. Instead, it should handling call chaining to the original `LoadLibraryA` and then calling an event callback. That way the event callback code is isolated from the call chaining, which is really just a detail of how that particular hook was implemented.

Further down the road, we could potentially allow for multiple event listeners, though I would suggest this is unneeded and out of scope for this issue.

PR #210, opened 2019-10-13
Extract `InstallTAppEventHooks()`

DanRStevens:
This supports issue #194. I'm taking a bit of a break, and this seemed like a good stopping point.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -95,7 +95,7 @@ bool InstallTAppEventHooks()
 	return true;
 }
 
-int TApp::Init()
+bool InstallDepPatch()
 {
 	// Set the execute flag on the DSEG section so DEP doesn't terminate the game
 	const std::size_t destinationBaseAddress = 0x00585000;
@@ -104,6 +104,14 @@ int TApp::Init()
 	if (!success) {
 		PostError("Error unprotecting memory at: 0x" + AddrToHexString(destinationBaseAddress));
 	}
+
+	return success;
+}
+
+int TApp::Init()
+{
+	// Install DEP patch so newer versions of Windows don't terminate the game
+	InstallDepPatch();
 
 	// Order of precedence for loading vol files is:
 	// ART_PATH (from console module), Console Module, Ini Modules, Addon directory, Game directory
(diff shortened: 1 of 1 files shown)


srcDLL/DllMain.cpp
```diff
@@ -60,11 +63,11 @@ BOOL WINAPI DllMain(HINSTANCE hInstance, DWORD dwReason, LPVOID reserved)
 		// Failure means op2ext.dll was loaded by something else, such as a unit test
 		// For unit tests, just stay in memory, it's not an error if this fails
 		if (EnableOp2MemoryPatching()) {
-			// Replace call to gTApp.Init with custom routine
-			// This hook is needed to further bootstrap the rest of module loading
+			// Initiall TApp event hooks
```
Brett208: Typo in comment. I think it should read "Initial"

srcDLL/DllMain.cpp
```diff
@@ -76,7 +79,23 @@ BOOL WINAPI DllMain(HINSTANCE hInstance, DWORD dwReason, LPVOID reserved)
...
+	// Replace call to gTApp.ShutDown with custom routine
+	Op2RelinkCall(0x004A88A5, GetMethodVoidPointer(&TApp::ShutDown));
+
+	// Replace call to LoadLibrary with custom routine (address is indirect)
+	Op2MemSetDword(loadLibraryDataAddr, &loadLibraryNewAddr);
+
+	return true;
+}
+
+bool InstallDepPatch()
```
Brett208: I don't really know what a DEP is or the DSEG segment. Might be worth explaining a bit more in comments?

---

It might be wise to put a declaration for the function InstallDepPatch at the top of the file in case it gets moved around. It would follow the current convention a bit better I think.

srcDLL/DllMain.cpp
```diff
@@ -161,10 +182,15 @@ HINSTANCE __stdcall LoadShell(LPCSTR lpLibFileName)
 
 	if (hInstance) // if good, then setup the language data and call the mod
```
Brett208: I think the comment associated with the if statement on hInstance would maybe best be deleted since we are not actually setting up language data. 

It could be moved as a comment on the function OnLoadShell since the work is now done there.

srcDLL/DllMain.cpp
DanRStevens: Oh geeze, I think I got confused between "initial" and "install". Thanks for pointing that out.

DanRStevens:
Ok, all very good points concerning the comments.

PR #211, opened 2019-10-14
Cleanup ini test setup

DanRStevens:
Test data is created and now deleted within the test that uses it.


DanRStevens:
We could potentially use the IniFile class. I was tempted to use that to do the writing, though I wanted to stick to existing code and keep it as close as possible.

Brett208:
I thought about that too and think it is a good idea.

PR #212, opened 2019-10-14
Load ini modules via ini section

### LINKED ISSUE #139: Allow loading external modules in the ini file via a full section

Brett208:
Closes #139.
Partially addresses #76.

The different tests for ModuleLoader and modules are spread across 3 test files now. I think it is good we are separating them out, but maybe some thought should be placed into which tests are run in which file to prevent duplicating work and easing organization. Maybe better for a separate branch?

### DIFF
--- a/srcStatic/ModuleLoader.cpp
+++ b/srcStatic/ModuleLoader.cpp
@@ -91,7 +91,7 @@ bool ModuleLoader::IsModuleRequested(const std::string& sectionName, const std::
 		return false;
 	}
 
-	PostError("Module named " + moduleName + " contains an innapropriate setting. It must be set to Yes or No");
+	PostError("Module named " + moduleName + " contains an innapropriate setting of " + isModuleRequested + ". It must be set to Yes or No");
 	return false;
 }
 
--- a/srcStatic/ModuleLoader.cpp
+++ b/srcStatic/ModuleLoader.cpp
@@ -25,7 +25,7 @@ ModuleLoader::ModuleLoader(IniFile iniFile, std::vector<std::string> consoleModu
 
 void ModuleLoader::RegisterBuiltInModules()
 {
-	if (IsBuiltInModuleRequested("IPDropDown")) {
+	if (IsModuleRequested("BuiltInModules", "IPDropDown")) {
 		RegisterModule(std::make_unique<IPDropDown>());
 	}
 }
@@ -80,9 +80,9 @@ void ModuleLoader::RegisterIniModules()
 	}
 }
 
-bool ModuleLoader::IsBuiltInModuleRequested(const std::string& moduleName)
+bool ModuleLoader::IsModuleRequested(const std::string& sectionName, const std::string& moduleName)
 {
-	const auto isModuleRequested = ToLower(iniFile.GetValue("BuiltInModules", moduleName));
+	const auto isModuleRequested = ToLower(iniFile.GetValue(sectionName, moduleName));
 
 	if (isModuleRequested == "yes") {
 		return true;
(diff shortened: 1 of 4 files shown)


srcStatic/ModuleLoader.cpp
```diff
@@ -67,22 +67,23 @@ void ModuleLoader::RegisterConsoleModules()
 
 void ModuleLoader::RegisterIniModules()
 {
-	const auto sectionNames = GetModuleNames("ExternalModules");
-
-	for (const auto& sectionName : sectionNames)
+	for (const auto moduleName : iniFile.GetKeyNames("ExternalModules"))
```
Brett208: I waffled between ExternalModules and IniModules. Sort of weird to have an IniModules section name in an ini file. Happy to change if @DanRStevens has preference one way or other?

test/IniModule.test.cpp
```diff
@@ -0,0 +1,53 @@
...
+		//iniFileStream << "Test = yes\n";
+	}
+
+	iniFileStream.close();
+}
+
+TEST(IniModule, NoDll)
+{
+	const auto iniFilename = fs::path(GetExeDirectory()) / fs::path("IniModuleTest.ini");
+	WriteExternalModuleIniFile(iniFilename.string(), { "Test = yes" });
```
Brett208: I'm not sure it is legal to pass into a string_view argument into a function without first recording the string to variable. Seems to work fine on MSVC though.

test/IniModule.test.cpp
```diff
@@ -0,0 +1,53 @@
...
+void WriteExternalModuleIniFile(std::string_view iniFilename, const std::vector<std::string_view> sectionPairs)
+{
+	std::ofstream iniFileStream(iniFilename);
+
+	iniFileStream << "[ExternalModules]\n";
+
+	for (const auto& sectionPair : sectionPairs)
+	{
+		iniFileStream << sectionPair << "\n";
+		//iniFileStream << "Test = yes\n";
```
DanRStevens: Maybe get rid of this commented out line.

test/IniModule.test.cpp
DanRStevens: Looks like the standard library packaged with Mingw doesn't have an overload taking `std::string_view` for the `std::ofstream` constructor. I guess we can just use a `const std::string&` for now.

srcStatic/ModuleLoader.cpp
DanRStevens: Uhh, I don't feel joy about these names, but I don't have a better suggestion right now. Maybe best to look at this again when I'm not tired. I'll open an issue to track.

test/IniModule.test.cpp
Brett208: Thanks for catching my sloppiness here.

test/IniModule.test.cpp
Brett208: Okay, the other option would be to define our own overload of ofstream for mingw that takes a string_view. :-) I know how to do that in C#, not sure what the C++ equivalent is without researching (I assume it is possible).

Thanks for fixing for me.

Brett208:
Thanks for the quick review. Yeah, it actually shortened the non unit test code base a bit and makes the ini file nicer to work with. The unit tests are lacking both because there is no way to test normal flow (no errors) and there is no real way to tell if the right errors are encountered in the 'bad' test cases. As mentioned, probably best for a different branch to deal with.

I made a slight improvement using std:endl after approval, but didn't think this was worth another review.

-Brett

test/IniModule.test.cpp
DanRStevens: We would have to add/modify things in the `std::` namespace, which creates upgrade and portability issues. Probably best to just leave this one alone.

It's also likely we'd need to do a `std::string_view` to `std::string` conversion to get it working, which kind of defeats the purpose of using `std::string_view` in the first place.

PR #213, opened 2019-10-14
Test logger

Brett208:
First take at creating a Test Logger. I think this could be used to test normal logging and error posting. Creation and cleanup within unit tests is bulky.


Brett208:
Unfortunately, I'm not yet versed in what Google Mock is really for. 

Perhaps we should delete LoggerMock and just go with this or maybe merge the two concepts? I like how this code is coming together.

-Brett

DanRStevens:
My preference is to stick with the Google Mock derived class. I believe it would be more feature complete (and standard) than anything we could hope to produce ourselves for a handful of tests. (That's actually one of the reasons they give for using Google Mock).

I should probably try my hand at it first though before judging. For one, the syntax will certainly be a bit different. It's not obvious how to use Google Mock objects without reading the documentation or seeing some example code. Maybe I can try writing up some sample code in a new branch.

PR #215, opened 2019-10-15
Load ini modules via ini section

Brett208:
Missed commits when I merged (didn't push local changes before merging on GitHub)


DanRStevens:
Heh, was about the comment about the lack of commits you had referenced. Looks good.

PR #217, opened 2019-10-16
Remove duplicated module tests

### LINKED ISSUE #137: Switch TestModule from reporting in output window to using logger
Currently the TestModule reports successful calling of different functions across DLL boundaries via the IDE's output window. That output is mixed with a bunch of other messages. 

I think it would be better behaviour to log the TestModule's finding with the logger for easier location and to exercise the logger itself.

Brett208:
Closes #137

I just left the portions of code that appear to be testing things that were not covered by existing unit tests. For example, checking that a module with a specific name is loaded.

I thought about fully removing this project, but it is still useful for testing beyond what the unit tests cover.

Example output from a console module test:

```
2019-10-16 02:01:36 UTC [op2mod.dll] Test Module mod_init called.
2019-10-16 02:01:36 UTC [op2mod.dll] Test Module mod_run called.
2019-10-16 02:01:36 UTC [op2mod.dll] Console Module TestModule is loaded
2019-10-16 02:01:36 UTC [op2mod.dll] Module TestModule is loaded
2019-10-16 02:01:36 UTC [op2mod.dll] The following 4 modules are loaded (ini and console combined):
2019-10-16 02:01:36 UTC [op2mod.dll]    TestModule
2019-10-16 02:01:36 UTC [op2mod.dll]    IPDropDown
2019-10-16 02:01:36 UTC [op2mod.dll]    NetFix
2019-10-16 02:01:36 UTC [op2mod.dll]    NetHelper
2019-10-16 02:01:38 UTC [op2mod.dll] Test Module mod_destroy called.
```


TestModule/TestFunctions.cpp
```diff
@@ -1,5 +1,4 @@
 #include "TestFunctions.h"
-#include "CountFilesByTypeInDirectory.h"
```
Brett208: This function has not been used within the code base for a while now. Do we want to keep it around / move it for future use or just toss it? I'm fine with either choice...

DanRStevens: I'm fine with just tossing it. It would still be a part of history if we ever wanted to resurrect it.

Actually, the same thing may have happened to the `ParseCsv` and associated `Split` and `Trim` code. The INI file updates that reads the keys from a section replaces their only use.

Brett208: The trim functions are fairly robust and we'll written. Tough to just toss them.

DanRStevens: I know, especially with all the time and thought I put into them. But they are still part of history, so easy to bring back if we ever need to.

Might be good to get rid of them so new users aren't confused by all that code that's never used.

Brett208: OK, deleted the function.

PR #218, opened 2019-10-16
Convert log message tests to use `LoggerMock`

DanRStevens:
Based on PR #213, and set to merge there.

A few notes:
- I added an empty test fixture class that inherits from `LogMessageTest`. This is simply to give a better name to the test in the `TEST_F` macro, and allow for better grouping.
- The `EXPECT_CALL` must be set before the expected call is triggered.
- The `_` (`::testing::_`) matches anything.

### DIFF
--- a/test/IniModule.test.cpp
+++ b/test/IniModule.test.cpp
@@ -17,9 +17,9 @@ protected:
 	}
 
 	// sectionPairs is a non-line terminated list of keys and values. EG: TestModule = No
-	void WriteExternalModuleIniFile(const std::string& iniFilename, const std::vector<std::string_view> sectionPairs)
+	void WriteExternalModuleIniFile(const std::vector<std::string_view> sectionPairs)
 	{
-		std::ofstream iniFileStream(iniFilename);
+		std::ofstream iniFileStream(iniFilename.string());
 
 		iniFileStream << "[ExternalModules]" << std::endl;
 		for (const auto& sectionPair : sectionPairs) {
@@ -33,7 +33,7 @@ protected:
 
 TEST_F(IniModuleTest, NoDll)
 {
-	WriteExternalModuleIniFile(iniFilename.string(), { "Test = yes" });
+	WriteExternalModuleIniFile({ "Test = yes" });
 
 	ModuleLoader moduleLoader(IniFile(iniFilename.string()), {});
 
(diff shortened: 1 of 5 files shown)


Brett208:
I'm having trouble understanding why this was based off of PR #213 instead of master. I don't think it is using any of PR #213?

DanRStevens:
It replaces the use of "TestLog.h" related code, introduced in that PR. This impacts calls to `ResetTestLog()`, and `GetErrorLogger()`, and the `EXPECT` calls that use them.

I didn't go so far as to actually remove the newly added class, just replace uses of it.

Brett208:
I've been reviewing this more. This appears to only be testing Logger::Log is called with the proper string.

In the original branch I was distinguishing in the mock object if normal log was called or if PostErrorMessage was called. If we were to switch to using GoogleMock, I would want to see this fidelity.

Also, having a Logger and Log class is proving difficult for me to distinguish between. I'd like to request renaming one or both to ease reading the code base.

-Brett

DanRStevens:
Ahh, right. I forgot that detail. Should be easy enough to add extra `LoggerMock` objects, install them in the `SetUp()` method, and remove them in the `TearDown()` method. Might need to update a few tests though, in case they generate warnings from unexpected logging calls that weren't previously intercepted.

Curiously, I noticed I was replacing calls to `GetErrorLogger()` code with regular `Log` message checks, and the tests were working as expected. Not sure what happened there. There were no uses of `GetLogger()`.

Brett208:
@DanRStevens, I think I could live with this instead of the custom mock object I had created. What do you think of the changes I pushed?

Thanks,
Brett

DanRStevens:
Well, it is a little annoying right now to double up the `EXPECT_CALL`. Long term, I see that part going away. Right now the `PostErrorMessage` contains an internal call to `Log`, which I would like to see removed. Instead, the double logging can be done with a log distributor class that contains references to the two logger objects. Such a setup will not affect unit testing. Since I assume the internal call to `Log` will eventually disappear, the doubled up `EXPECT_CALL` should also eventually disappear. As such, I'm maybe not so fond of the changes that wrap the two calls into a helper function.

DanRStevens:
Merging of other branches was creating quite a bit of noise here. To avoid seeming like there were a really large number of commits on this branch, I merged an updated master to the base branch (PR #213), and then rebased this work on that.

DanRStevens:
By default Google Mock will warn for any unexpected calls.

PR #220, opened 2019-10-17
Remove Main.cpp from unit test project

DanRStevens:
The file is basically empty now anyway.


DanRStevens:
Mmm, maybe? Though I'm kind of hoping not. I much prefer when unit test code is self contained. Less chance for one test to interfere with another test.

PR #222, opened 2019-10-17
Remove dead code

### LINKED ISSUE #219: Remove dead code
In PR #217 it was noted there was some dead code in the TestModule project. Namely:
`CountFilesByTypeInDirectory`

Another point was the `ParseCsv`, `Split`, and `Trim` methods, which are also now dead code. These methods were used to support parsing of INI file settings. However, INI module loading was changed to read all keys from a special section, so no longer needs to parse a command separated value list.

DanRStevens:
This closes #219.

We can easily revert this change in the future, should we ever decide to resurrect the code.

Affected functions (included for easy search):
- `Trim`
- `TrimFront`
- `TrimBack`
- `Split`
- `SplitAndTrim`
- `ParseCsv`


DanRStevens:
It kind of kills me to do it, ... but yes. (So long as it's retained in history).

PR #224, opened 2019-10-17
Test module header includes

DanRStevens:
Noticed a couple of small things that needed fixing.

### DIFF
--- a/TestModule/DllMain.cpp
+++ b/TestModule/DllMain.cpp
@@ -17,9 +17,6 @@ Initialized = "True"
 
 #include "TestFunctions.h"
 #include "op2ext.h"
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#include <string>
 
 //Test external interface is compatible with C compiler.
 #include "CompatibilityTest.c"
--- a/TestModule/DllMain.cpp
+++ b/TestModule/DllMain.cpp
@@ -18,8 +18,6 @@ Initialized = "True"
 #include "TestFunctions.h"
 #include "op2ext.h"
 
-//Test external interface is compatible with C compiler.
-#include "CompatibilityTest.c"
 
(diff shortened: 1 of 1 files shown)


TestModule/DllMain.cpp
```diff
@@ -17,12 +17,7 @@ Initialized = "True"
 
 #include "TestFunctions.h"
 #include "op2ext.h"
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#include <string>
 
-//Test external interface is compatible with C compiler.
-#include "CompatibilityTest.c"
```
Brett208: How do we guarantee that MSVC will compile this file if it is not included?

DanRStevens: It will be since it's already listed as a source file for that project:
```
<ClCompile Include="CompatibilityTest.c" />
```

On AppVeyor for the build above, it can be seen building in the lines:
```
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x86\CL.exe /c /I../srcDLL /I"C:\Tools\vcpkg\installed\x86-windows\include" /Zi /nologo /W3 /WX- /diagnostics:classic /sdl /O2 /Oi /Oy- /GL /D _WINDLL /D _MBCS /Gm- /EHsc /MD /GS /Gy /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /std:c++17 /Fo"Release\\" /Fd"Release\vc141.pdb" /Gd /TC /analyze- /FC /errorReport:queue CompatibilityTest.c
  CompatibilityTest.c
```

This is after the change that removed the `#include` line.

Brett208: Oh, ok. I see this locally as well. For some reason, I thought it wasn't building the file back when I wrote the TestModule.

PR #226, opened 2019-10-17
Organize TApp event hook code

DanRStevens:
This is cleanup to support Issue #194. The `TApp` related code still hasn't been moved to it's own file, but has been nicely organized together at the bottom of DllMain.cpp. Some decisions will need to be made concerning events, and event handler visibility before the code can be moved out. For now this seems like a good stopping point.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -129,11 +129,6 @@ void OnInit()
 	volList->LoadVolFiles();
 }
 
-void OnShutdown()
-{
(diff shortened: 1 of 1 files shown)


DanRStevens:
Oops, my mistake. It seems the `__stdcall` was actually needed. For some reason I thought that was just specifying a default. Turns out the default (`__cdecl`) is to have the caller cleanup parameters on the stack, while we need the callee to cleanup parameters on the stack.

Issue #93, opened 2019-05-04
Strip directory separators when forming ConsoleModuleName


Brett208:
The console module name is formed from the /loadmod argument. This could be passed a directory with different path specific decorations that should be stripped to allow accepting a wider range of options without undefined behaviour within op2ext.

From @DanRStevens:

In terms of automatically handling conversions of weird paths with ".", "..", and normalizing trailing slash, perhaps the following methods may be helpful:

- https://en.cppreference.com/w/cpp/filesystem/canonical
- https://en.cppreference.com/w/cpp/filesystem/relative
- https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal

These methods should provide a way to normalize paths that use "./" or "../". Some of them may also normalize use of symbolic links (which maybe we do or don't want to do). They will probably also help normalize trailing slash.

Looking at the first two, I think what you might want is fs::relative(moduleFolderPath, executableFolder), with any trailing slash stripped (which I think might be the default behavior).

Looking at the last method, you might use moduleFolderPath.lexically_relative(executableFolder). I believe the difference is the "lexical" method works on the path string alone, without ever checking the filesystem, and hence would not be bothered by path elements that don't exist and would not be able to resolve symbolic links.

DanRStevens:
Tagging PR #91 as the source of the discussion that led to this issue.

Re-reading the discussion, it seems to suggest an approach where we allow the user to specify console modules by using a path to the folder, in whatever manner they choose, and then normalize it to a well known form. The normalized name allows for things such as detecting duplicate module names.

In contrast, the discussion does not seem to suggest locking down the user supplied portion so they can only specify the module name in one well known format.

There appears to be no restriction on multiple levels of subfolders to reach the module folder. Currently, I see no reason to impose an artificial restriction such as a single folder level. If nested folders are used, the whole relative path, including any directory separators, will become the module name.

DanRStevens:
It seems the `<experimental/filesystem>` library included with Mingw does not include:
`relative`, `path::lexically_normal`, `path::lexically_relative`

This leaves only `canonical` for transforming and normalizing paths.

Note that `canonical` throws errors for paths that don't exist.

The `canonical` method does not take a base path, like the relative methods do, so paths would be absolute, rather than relative to some fixed directory such as the game directory.

DanRStevens:
After some thought, I'm wondering if this is something we need to fix, or if we should perhaps close this as a non-issue.

[...]

PR #230, opened 2019-10-20
Make locate vol files use an absoluate directory

Brett208:
I would recommend reviewing and merging this branch before #227.

I think it will ensure directories provided by ini modules will be correct.


DanRStevens:
The `fs::relative` method is not available in the `<experimental/filesystem>` library shipped with Mingw. I actually just discovered that earlier today, and commented about it in issue #93.

Brett208:
I've never used filesystem::relative before. It worked like I was hoping on MSVC. Not sure, but maybe, mingw just doesn't actually support all of filesystem yet?

DanRStevens:
With that said though, I have been wanting to update this method for a while now. It can probably even be moved to the static library project. Though I think I would like to remove the exception handling and let the caller deal with any exceptions.

srcDLL/DllMain.cpp
```diff
@@ -98,13 +98,15 @@ void OnInit()
 
 	// Find VOL files from additional folders
 	for (std::size_t i = 0; i < moduleLoader->Count(); ++i) {
-		std::string moduleDirectory = moduleLoader->GetModuleDirectory(i);
+		const auto moduleDirectory = moduleLoader->GetModuleDirectory(i);
```
DanRStevens: Maybe best to use `const auto&`.

srcDLL/DllMain.cpp
```diff
@@ -98,13 +98,15 @@ void OnInit()
...
 		if (!moduleDirectory.empty()) {
-			LocateVolFiles(moduleLoader->GetModuleName(i));
+			LocateVolFiles(moduleDirectory);
 		}
 	}
-	LocateVolFiles("Addon");
-	LocateVolFiles(); //Searches root directory
+
+	const auto exeDirectory = fs::path(GetExeDirectory());
+	LocateVolFiles((exeDirectory / fs::path("Addon")).string());
```
DanRStevens: No need for the `fs::path` wrapper. A `std::string` or `const char*` will work just as well with `operator /`.

srcDLL/DllMain.cpp
```diff
@@ -143,12 +141,13 @@ void LocateVolFiles(const std::string& relativeDirectory)
...
 			if (extension == ".vol") {
-				volList->AddVolFile((fs::path(relativeDirectory) / filePath.filename()).string());
+				auto relativePath = fs::relative(absoluteDirectory, GetExeDirectory()) / filePath.filename();
+				volList->AddVolFile(relativePath.string());
 			}
 		}
 	}
 	catch (const fs::filesystem_error& e) {
-		Log(e.what());
+		Log("Unable to locate vol files in provided directory " + absoluteDirectory + ". " + std::string(e.what()));
```
DanRStevens: This message only contains information available to the caller. Perhaps the caller should catch and log exceptions.

Brett208:
I'm taking a break if you want to play with this. 

I'd propose writing a function like IsDirectory that mimicks fs::relative. That way when mingw pulls in full support for filesystem, we can delete the function and just use fs::relative. 

I was assuming that mingw was no longer using an experimental filesystem for some reason...

srcDLL/DllMain.cpp
```diff
@@ -124,13 +126,9 @@ void OnShutdown()
 /**
 Prepares all vol files found within the supplied relative directory from the Outpost 2 executable
 for inclusion in Outpost 2. Does not recursively search subdirectories.
-
-@param relativeDirectory A directory relative to the Outpost 2 exectuable. Default value is an empty string.
 */
-void LocateVolFiles(const std::string& relativeDirectory)
+void LocateVolFiles(const std::string& absoluteDirectory)
 {
-	const auto absoluteDirectory = (fs::path(GetExeDirectory()) / relativeDirectory).string();
```
DanRStevens: I was thinking perhaps a `basePath` argument could be added. Paths could be given relative to the `basePath`. We could offer a one parameter overload that uses `GetExeDirectory()` for the `basePath`.

srcDLL/DllMain.cpp
```diff
@@ -143,12 +141,13 @@ void LocateVolFiles(const std::string& relativeDirectory)
 			const auto extension = ToLower(filePath.extension().string());
 
 			if (extension == ".vol") {
```
DanRStevens: We could generalize this method by making the desired extension a parameter.

srcDLL/DllMain.cpp
```diff
@@ -143,12 +141,13 @@ void LocateVolFiles(const std::string& relativeDirectory)
...
-				volList->AddVolFile((fs::path(relativeDirectory) / filePath.filename()).string());
+				auto relativePath = fs::relative(absoluteDirectory, GetExeDirectory()) / filePath.filename();
+				volList->AddVolFile(relativePath.string());
```
DanRStevens: We could generalize this method by using some sort of callback for what to do with the found path.

srcDLL/DllMain.cpp
Brett208: In retrospect, this is probably best not being a reference. There is no guarantee a module doesn't just generate the directory and return it without storing a local copy.

srcDLL/DllMain.cpp
Brett208: Changing it to read:
```C++
LocateVolFiles(GetExeDirectory() / "Addon");
```

Produces: 
E0349	no operator "/" matches these operands	op2extDLL	\op2ext\srcDLL\DllMain.cpp	108

Perhaps the operator overload for path concatenation is mingw specific?

srcDLL/DllMain.cpp
Brett208: I suspect this branch is going to balloon in scope...

srcDLL/DllMain.cpp
Brett208: See if the function I placed in the FileSystem library satisfies your intention. It sounds like you want to go a step further and act on results instead of return a container. I would accomplish maybe by passing a function as an argument or something... Not sure I want to go that route right now due to technical difficulty accomplishing on my part.

srcDLL/DllMain.cpp
Brett208: I moved this function to within VolList. I think catching the exception there makes sense. Let me know if you still want it more generalized.

srcDLL/DllMain.cpp
Brett208: What about:

```C++
AddVolFilesFromAbsoluteDirectory(const std::string& absoluteDirectory);

// relativeDirectory is from the directory containing the Outpost 2 exectuable
AddVolFilesFromRelativeDirectory(const std::string& relativeDirectory);
```

Brett208:
I wanted to write some basic unit tests for VolList since I added a new public function to it. However, we need to get the mock test logger branches figured out first to be actually productive, so I just left it as a stub. 

Maybe better to write an issue to provide some VolList tests.

The fs::relative not on mingw is still an issue. If recreating fs::relative seems difficult, we could just implement an alternate solution and add a comment saying to switch to relative when mingw catches up with C++17 standard.

-Brett

srcDLL/DllMain.cpp
DanRStevens: C++ has lifetime extension rules to handle cases like that. It's not a problem. A returned temporary object will not go out of scope and be destructed until after the capturing reference goes out of scope.

srcDLL/DllMain.cpp
DanRStevens: Yes, correct. The line above originally read `fs::path(GetExeDirectory())`. That wrapping is enough. The `operator/` is overloaded to take both `operator/(fs::path, fs::path)` and `operator/(fs::path, std::string)`. It does not however take `operator/(std::string, std::string)`. As long as the first parameter is wrapped in `fs::path` it will work.

srcDLL/DllMain.cpp
DanRStevens: Hmm, it seems my original idea here was a bit fuzzy.

I was thinking for a generalized function:
```cpp
ResultType FindFilesInFolder(const std::string& relativeDirectory, const std::string& basePath) {
  // Main implementation here
}

ResultType FindFilesInFolder(const std::string& relativeDirectory) {
  return FindFilesInFolder(relativeDirectory, GetExeDirectory());
}
```

The resulting paths would be relative to the `basePath` parameter.

The above ignores the details of how the paths are returned. Maybe through a `std::vector`, or maybe to a callback, which might be passed as an extra parameter. It also ignores other parameters that might be added for a generalized function, such as a filename extension to look for.

----

Whether we can form of VOL list from relative or absolute folders is a concern I had not considered. I don't even really remember why the system was designed the way it was to only take relative paths.

srcDLL/DllMain.cpp
DanRStevens: I kind of do that don't I. Perhaps I should have pushed some of those thoughts into an issue.

srcDLL/DllMain.cpp
Brett208: OK, I didn't realize that was the case. I think the tip of the weeks would not recommend implementing this. Although, now that I know it is legal, I guess I don't care too much one way or the other. 

From https://abseil.io/tips/101

> Scenario #8 is what allows most reflexive use of references (that is “Oh, I don’t want to copy so I’ll just assign into a reference” without necessarily thinking about what is being returned). However, because of #1, it’s also not really doing anything for you: there probably wouldn’t have been a copy in the first place. Further, now readers of your code have to contend with your local variable being of type const string& instead of string, and thus worry about whether the underlying string has gone out of scope or changed.

srcDLL/DllMain.cpp
Brett208: PR#232 is addressing this.

srcDLL/DllMain.cpp
Brett208: I believe Outpost 2 uses relative paths to the vol files internally when loading.

srcStatic/FileSystemHelper.cpp
```diff
@@ -40,3 +41,29 @@ bool Exists(const std::string& path)
...
+
+std::vector<std::string> LocateFilesWithExtension(const std::string& directory, std::string extension)
+{
+	if (!IsDirectory(directory)) {
+		return {};
+	}
+
+	if (extension.empty()) {
+		throw std::runtime_error("provided extension cannot be empty");
+	}
```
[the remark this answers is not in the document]
Brett208: I'm fine with allowing empty extensions. For this function to be proper, I believe each directory entry needs to be checked to make sure it is a file. Otherwise an empty extension might return all the subdirectories. Also, a directory named bad.vol might pass right now.

Since this function is changed by another branch, I'm going to hold off working on it for now.

Brett208:
I'm find with using the word Find instead of Locate (accomplished in separate branch). Work on this branch should be on hold until PR #232 is finished due to how intertwined they are.

srcStatic/FileSystemHelper.cpp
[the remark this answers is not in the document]
DanRStevens: Empty extensions doesn't have anything to do with file versus directory. They are orthogonal concerns.

You do raise a valid point about not wanting to pass a directory name to be loaded as a VOL file.

DanRStevens:
We may have broken console module loading. I tried testing with `/loadmod cm1`. It ran without error, but the color module clearly wasn't loaded. I'll look more into this shortly.

Brett208:
I just checked, and this branch is properly loading a console module and an ini module on my computer. I did notice that loose resources stored in a console module directory were not being recognized. It appears directories must be passed as absolute, which I updated in a commit.

@DanRStevens, could you double check and see if you can load modules on your build? If so, I'm okay with merging unless you want to write up some unit tests for the new Filesystem function beforehand.

-Brett

srcStatic/ResourceSearchPath.cpp
```diff
@@ -12,7 +12,7 @@ class ResManager {
 	bool GetFilePath(const char* resourceName, /* [out] */ char* filePath) const;
 };
 
-
+// All paths must be an absolute directory
```
DanRStevens: Hmm, the `ResourceSearchPath` code has no technical requirement for absolute paths. This does seem to get to the root of the problem though, in that relative paths are relative to the current working directory, and depending on how the game was started, that might not be the executable folder.

I'd recommend adding a bit more detail to this comment.

PR #232, opened 2019-10-21
Updates for pr230

DanRStevens:
Here are the suggestions I made for PR #230. They compile on Linux.

My initial testing shows VOL loading seems to be broken. I haven't had time to investigate the cause.

### DIFF
--- a/srcStatic/FileSystemHelper.cpp
+++ b/srcStatic/FileSystemHelper.cpp
@@ -48,10 +48,6 @@ std::vector<std::string> LocateFilesWithExtension(const std::string& directory, 
 		return {};
 	}
 
-	if (extension.empty()) {
-		throw std::runtime_error("provided extension cannot be empty");
-	}
-
 	ToLowerInPlace(extension);
 	std::vector<std::string> filesWithExtension;
 
--- a/srcStatic/FileSystemHelper.cpp
+++ b/srcStatic/FileSystemHelper.cpp
@@ -42,9 +42,11 @@ bool Exists(const std::string& path)
 	return fs::exists(modifiedPath);
 }
 
-std::vector<std::string> FindFilesWithExtension(const std::string& directory, std::string extension)
+std::vector<std::string> FindFilesWithExtension(const std::string& basePath, const std::string& relativeSearchPath, std::string extension)
 {
-	if (!IsDirectory(directory)) {
+	auto directory = fs::path(basePath) / relativeSearchPath;
+
+	if (!IsDirectory(directory.string())) {
 		return {};
 	}
 
@@ -57,7 +59,7 @@ std::vector<std::string> FindFilesWithExtension(const std::string& directory, st
 		const auto extensionToCheck = ToLower(filePath.extension().string());
 
 		if (extensionToCheck == extension) {
-			filesWithExtension.push_back(filePath.string());
+			filesWithExtension.push_back(filePath.string().substr(basePath.size()));
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


DanRStevens:
Found a fix for the VOL loading. The problem was from trying to concatenate two absolute paths together. For now I went with having `ConsoleModule::Directory` return a relative path.

srcStatic/FileSystemHelper.cpp
```diff
@@ -42,14 +42,12 @@ bool Exists(const std::string& path)
...
 {
-	if (!IsDirectory(directory)) {
-		return {};
-	}
+	auto directory = fs::path(basePath) / relativeSearchPath;
 
-	if (extension.empty()) {
-		throw std::runtime_error("provided extension cannot be empty");
+	if (!IsDirectory(directory.string())) {
+		return {};
```
Brett208: Would it be better to throw an exception here to indicate that improper data was provided?

srcStatic/GameModules/ConsoleModule.cpp
```diff
@@ -8,14 +8,15 @@
 
 ConsoleModule::ConsoleModule(const std::string& moduleName) : DllModule(moduleName)
 {
-	moduleDirectory = (fs::path(GetExeDirectory()) / moduleName).string() + "\\";
+	moduleDirectory = moduleName + "\\";
```
Brett208: Maybe this class level variable should be renamed relativeModuleDirectory

srcStatic/VolList.cpp
```diff
@@ -12,18 +12,17 @@ void VolList::AddVolFile(std::string volPath)
 	volPaths.push_back(std::move(volPath));
 }
 
-void VolList::AddVolFilesFromDirectory(const std::string& absoluteDirectory)
+void VolList::AddVolFilesFromDirectory(const std::string& relativeSearchPath)
```
Brett208: Why did you go with the term SearchPath for the argument? I think a path could be either a directory or a filename, but we specifically want a relativeDirectory?

Brett208:
I tried to use the GitHub built in merge conflict resolver again after having not used it in a long time. It bites me every time and should still be avoided.

srcStatic/FileSystemHelper.cpp
DanRStevens: Technically that aspect isn't really addressed in this PR. It just shows up as a highlighted change due to changes to surrounding lines. The diff is a bit confused here.

I'm fine with either way. I think I have a slight preference for returning an empty list (as the code previously did) rather than throwing an exception. If you try to find files in a non-existent directory, there are of course no such files to be found.

srcStatic/GameModules/ConsoleModule.cpp
DanRStevens: I'm fine either way. I went with the convention that `Directory()` returns a relative path, so things kind of match here.

srcStatic/VolList.cpp
DanRStevens: You're right. I'm used to using APIs where you can specify a search pattern. In that case the pattern is not necessarily a directory name. But with the C++ filesystem API, it must be a directory name.

srcStatic/FileSystemHelper.cpp
Brett208: Ok, let's just return empty for non-existent directory.

DanRStevens:
Ok, I'm done with this for now. I need to run. If you want to rename `moduleDirectory`, that's up to you.

PR #234, opened 2019-10-21
Unify Detection of Functions Exported by Modules

Brett208:
I think this breaks NetFix from loading but I haven't figured out why yet.

### DIFF
--- a/srcStatic/GameModules/DllModule.cpp
+++ b/srcStatic/GameModules/DllModule.cpp
@@ -1,4 +1,5 @@
 #include "DllModule.h"
+#include "../Log.h"
 #include "../WindowsErrorCode.h"
 #include <stdexcept>
 
@@ -60,6 +61,9 @@ bool DllModule::Unload()
 	}
 	else if (unloadModuleFunctionConsole != nullptr) {
 		success = unloadModuleFunctionConsole();
+		if (!success) {
+			Log("Module reports error during unload: " + Name());
+		}
 	}
 
 	if (moduleDllHandle != nullptr) {
--- a/srcStatic/GameModules/DllModule.cpp
+++ b/srcStatic/GameModules/DllModule.cpp
@@ -2,6 +2,7 @@
 #include "../WindowsErrorCode.h"
 #include <stdexcept>
 
+
 DllModule::DllModule(const std::string& moduleName)
 	: GameModule(moduleName) { }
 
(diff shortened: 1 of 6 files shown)


srcStatic/GameModules/DllModule.cpp
```diff
@@ -18,4 +19,56 @@ void DllModule::LoadModuleDll(const std::string& dllPath)
...
+void DllModule::DetectExportedModuleFunctions()
+{
+	loadModuleFunctionIni = (LoadModuleFunctionIni)GetProcAddress(moduleDllHandle, "InitMod");
+	if (loadModuleFunctionIni == nullptr) {
+		loadModuleFunctionConsole = (LoadModuleFunctionConsole)GetProcAddress(moduleDllHandle, "mod_init");
+	}
+
+	unloadModuleFunction = (UnloadModuleFunction)GetProcAddress(moduleDllHandle, "DestroyMod");
+	if (unloadModuleFunction == nullptr) {
+		unloadModuleFunction = (UnloadModuleFunction)GetProcAddress(moduleDllHandle, "mod_destroy");
```
DanRStevens: These methods differ by the return type. The `DestroyMod` address should not be casted to a type with a non-void return type. It may cause random garbage values left in registers to be interpreted as a return type. This may fool calling code into thinking the module failed to unload correctly.

srcStatic/GameModules/DllModule.cpp
```diff
@@ -18,4 +19,56 @@ void DllModule::LoadModuleDll(const std::string& dllPath)
...
+		loadModuleFunctionConsole();
+	}
+}
+
+bool DllModule::Unload()
+{
+	bool success = true;
+
+	if (unloadModuleFunction != nullptr) {
+		success = unloadModuleFunction();
```
DanRStevens: Not all modules return a success value during unload.

We should probably log unload errors. (Console modules only).

srcStatic/GameModules/DllModule.cpp
```diff
@@ -18,4 +19,56 @@ void DllModule::LoadModuleDll(const std::string& dllPath)
...
+	}
+
+	if (moduleDllHandle != nullptr) {
+		FreeLibrary(moduleDllHandle);
+	}
+
+	return success;
```
DanRStevens: This return value is ignored. Perhaps we should consider getting rid of it. The best we can really do with a module that fails to unload is to log the error.

Brett208: good point about the difference between the two unload functions.

What do you think if we just change the specification for a console module to have a void return? Then they would both read:

```C++
typedef void(*UnloadModuleFunction)();
```

I just tested, and it looks like the cm1 module loads fine as a console module with the code tweaked. I could just pull out the success checks altogether.

Brett208:
I just retested this closely. It loads console and ini modules properly. Not sure what I was seeing the other day when I wrote it besides being a bit rushed for time maybe.

srcStatic/GameModules/DllModule.cpp
DanRStevens: It is tempting. I'd thought of doing that before. It should be safe to change the return type from `bool` to `void`.

I noticed the NetHelper module can actually return a failure result there.

I think I have a slight preference for retaining the official export signature as it was, and then just swallowing the result value in the `DllModule::Unload` method, with a possible log message in case of failure.

Ideally, modules should do their own error reporting, as it would be more detailed, however, older existing modules wouldn't have really had that option.

DanRStevens: I spoke with Arklon about this. It seems the return value was added as a way to potentially support a future in-game mod manager that could handle hot loading and unloading of modules, without a game restart. Potentially some mods would be tied to specific missions, such as the survivor missions.

Brett208: OK, I'm good with adding both signatures. I'll have it check for the bool version first, and if it isn't present, pull the void return version.

Brett208:
OK, I think it is ready. I used the TestModule to make sure console modules still loaded properly. NetFix seemed to load okay on the ini module side.

srcStatic/GameModules/DllModule.h
```diff
@@ -26,14 +26,16 @@ class DllModule : public GameModule
 	// The exported naming schema is different between console and ini functions
 	typedef void(*LoadModuleFunctionIni)(const char* iniSectionName);
 	typedef void(*LoadModuleFunctionConsole)();
-	typedef bool(*UnloadModuleFunction)();
+	typedef bool(*UnloadModuleFunctionIni)();
+	typedef void(*UnloadModuleFunctionConsole)();
```
DanRStevens: I believe these two types are reversed. The console module was the one with the `bool` return type.

Oh, and we should probably switch to `using` rather than `typedef`. They read in a more natural order.

PR #235, opened 2019-10-22
Add `LoggerDistributor` class

DanRStevens:
Rather than talk about this for ages, I thought I'd get off my butt and implement it.

This may impact PR #218.

### DIFF
--- a/srcStatic/LoggerDistributor.h
+++ b/srcStatic/LoggerDistributor.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#include "Logger.h"
+#include <vector>
+
+
+class LoggerDistributor : public Logger
+{
+public:
+	LoggerDistributor(std::vector<Logger*> loggers);
+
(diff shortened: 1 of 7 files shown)


srcStatic/LoggerDistributor.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "Logger.h"
+#include <vector>
+
+
+class LoggerDistributor : public Logger
+{
+public:
+	LoggerDistributor(std::vector<Logger*> loggers);
```
Brett208: What do you think about adding a comment to the constructor explaining how ownership of the loggers is supposed to work? I think these are more reference than pointers.

DanRStevens: Probably a good idea to document (non-)ownership characteristics, and non-nullability. I had originally tried to use references, but found that created a problem for vectors.

PR #244, opened 2019-10-27
Update logging code

### LINKED ISSUE #170: Review logging code
At some point the logging code should be reviewed, and perhaps made to be more regular.

Issue #155 and PR #162 made some changes to collect public logging and reporting functions into a common file. Included was some discussion of future direction.

It was planned to have global methods for each logging level:
```
void Log(...);
void LogError(...);
void LogDebug(...);
```

The current implementation looks more like:
```
void Log(const std::string& message, const std::string& moduleName = "op2ext.dll");
void PostErrorMessage(const std::string& errorMessage, const std::string& sourcePathFilename, long lineInSourceCode);
#define PostError(errorMessage) PostErrorMessage(errorMessage, __FILE__, __LINE__)
void LogDebug(const std::string& message);
```

Currently `LogError` is roughly implemented as the macro `PostError`, with the details in `PostErrorMessage`. We should try to make these consistently named. It would also be nice if we could somehow get rid of the macro irregularity. In particular, the macro requires a bit of naming irregularity since the macro name must be distinct from the underlying function name that it delegates to.

The parameters to these methods are currently different. Ideally these should be made more regular.

The `LogError` function takes a `__FILE__` and `__LINE__` location, while the others do not. By defaulting that with the macro, the remaining parameters are left more consistent. This does result in the macro irregularity however. We could also add `__FILE__` and `__LINE__` to the other methods to make them more consistent.

Additionally, the core `Log` method takes a `moduleName` parameter, which is absent from the other ones. We could perhaps push the `moduleName` detail and associated formatting up into the op2ext.cpp exported method. Direct calls to logging from within op2ext would then have a shorter parameter list. This would also free internal calls from the associated `moduleName` string formatting, which may or may not be a good thing.

The default `moduleName` parameter is set to `op2ext.dll`. This makes the code less generic and less re-usable in other projects. Moving that detail up into the op2ext.dll exported module interface could leave us with more generic and re-usable logging code.

C++20 will introduce [`source_location`](https://en.cppreference.com/w/cpp/utility/source_location). That could eventually be used to replace the `__FILE__` and `__LINE__` macro uses. The new functionality allows for much easier defaulting of the source location parameter, without having to use a macro wrapper.

We should create a logger distributor class, which can handle distributing log messages to multiple output logger destinations. This would enable a single object to output to both on screen sources, and to a log file, without having to tie the two functionalities too closely together.

The `OutputDebugString` code should be extracted into a `Logger` derived class. This would increase flexibility for debug logging.

We should add the ability to set the logger for each logging level. Currently we are missing this for the debug logging level:
```
void SetLogger(Logger* newLogger);
void SetLoggerError(Logger* newLogger);
void SetLoggerDebug(Logger* newLogger);
```

In DllMain.cpp there are multiple direct calls to the Win32 API method `OutputDebugString`. We should redirect these to `LogDebug`.

Related, though this might really be a separate issue:
The TestModule also makes direct use of the Win32 API method `OutputDebugString`.
We should review code for appropriate amounts of logging output. In particular, module load/run/unload events.
We should reduce/eliminate use of modal pop-up message boxes. (Non-modal would be much better).

DanRStevens:
This closes #170.

### DIFF
--- a/test/LogMessageTest.h
+++ b/test/LogMessageTest.h
@@ -8,7 +8,7 @@ public:
 	LoggerMock();
 	virtual ~LoggerMock();
 
-	MOCK_METHOD2(Log, void(const std::string& message, const std::string& moduleName));
+	MOCK_METHOD1(Log, void(const std::string& message));
 };
 
 
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -110,7 +110,7 @@ OP2EXT_API void Log(const char* message)
 	// These optimizations are however extremely unlikely when making
 	// calls across a module boundary (such as to exported methods).
 
-	Log(message, FindModuleName(_ReturnAddress()));
+	Log(FormatLogMessage(message, FindModuleName(_ReturnAddress())));
 }
 
 
--- a/srcStatic/Log.cpp
+++ b/srcStatic/Log.cpp
@@ -39,11 +39,11 @@ void SetLoggerDebug(Logger* newLogger) {
 
 
 // Output log message at standard logging level
-void Log(const std::string& message, const std::string& moduleName) {
+void Log(const std::string& message) {
 	// Make sure a logger has been set first
 	if (logger) {
 		// Delegate to internal logger
-		logger->Log(message, moduleName);
+		logger->Log(message);
 	}
 }
 
--- a/srcStatic/Log.h
+++ b/srcStatic/Log.h
@@ -11,7 +11,7 @@ void SetLogger(Logger* newLogger);
 void SetLoggerError(Logger* newLogger);
 void SetLoggerDebug(Logger* newLogger);
 
-void Log(const std::string& message, const std::string& moduleName = "op2ext.dll");
+void Log(const std::string& message);
 
 
 // Outputs a debug string. Output is not logged by the logger.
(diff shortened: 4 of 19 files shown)


test/LogMessageTest.h
```diff
@@ -8,7 +8,7 @@ class LoggerMock : public Logger {
 	LoggerMock();
 	virtual ~LoggerMock();
 
-	MOCK_METHOD2(Log, void(const std::string& message, const std::string& moduleName));
+	MOCK_METHOD1(Log, void(const std::string& message));
```
Brett208: I just realized the number at the end of MOCK_METHOD is the number of arguments a function is supposed to take.

DanRStevens: Yes. That was also the stumbling block I hit when I first tried writing a few Google Mock tests. The newer versions of Google Mock don't need that. Instead they separate out the return type from the argument types, and are somehow able to detect the number of parameters. The newer code samples use an updated macro.

DanRStevens:
Agreed about not liking the `FormatLogMessage` calls when calling `Log`. That's part of the reason why I suggested removing them in issue #245. Most uses of it don't seem particularly relevant. It was maintained here to keep behaviour the same so this PR could be a simple refactor with no visible runtime changes.

PR #248, opened 2019-10-27
Remove global

DanRStevens:
Noticed this and thought it was a bit odd.


DanRStevens:
Yes. The address is of a formatted text string. That's why the `_snprintf_s` call was needed to format a string from integer values before copying it.

Issue #239, opened 2019-10-24
Improve encapsulation of `VolList` updates


DanRStevens:
Currently the module loader locks the VOL list at a certain point in time. Any attempts to load a VOL list after some initialization point will fail. We could update the VOL list code to allow for updates at largely arbitrary times.

Currently the `VolList` class takes VOL filenames one at a time, and builds a list. At some point that list is finalized, translated to an internal game data structure, and the game is patched to use it. Instead, we can move the list building external to the class (it's really just pushing strings into a vector), and have the `VolList` class take an array of VOL filenames. During construction, the list would be used to initialize a private const `std::vector`, and build an associated internal game structure. A separate method, callable at a later time could be used to `Activate()` the list, which would patch the game to use it. This means we could have separate objects, corresponding to different VOL lists, and then swap them out at runtime, whenever we wanted by calling `Activate()`.

Threading note: Assuming only the main thread does resource lookup, then we can swap out whenever we want on the main thread. If multiple threads do resource lookups, then you'd want to be careful about swapping out the lists while another thread is doing a resource lookup. That might require some patching to add synchronization code. As far as I know, the game does all resource lookup from the main thread.

Proposed `VolList` public methods:
```cpp
VolList(std::vector<std::string> volPaths);
void Active();
```

----

Related:
- The exported API method to add VOL files would need to be updated
- We may need to add an exported API method to remove VOL files

DanRStevens:
Re-reading the notes, it seems the focus here is on support code changes involving `VolList`. I've updated the title to match.

There may need to be some additional hooking/patching required to changed the loaded VOLs after the game has loaded them. I will consider that out of scope for this issue. Similarly, adding a new method to the exported API to remove VOL files is also out of scope here. We can handle those items in a future issue.

PR #253, opened 2019-10-28
Use `vector` for `AddVolToList`

DanRStevens:
This is a bit of background work for issue #239. This collects all uses of `VolList` into the `OnInit()` method in DllMain.cpp.

The longer term plan is to rework `OnInit()` to build the full VOL list first, and then pass the full list to a new `VolList` constructor. Currently the code is still constructing an empty `VolList` and then adding VOL files one at a time.

I'm stopping here to come up with a plan for the next part. The VOL list building could be done in one of many ways, and all of them are a bit awkward with current C++ features.


DanRStevens:
I know what you mean. I was tempted to not push it until I'd come up with a plan for the rest of it.

PR #256, opened 2019-10-31
Use `std::size_t` instead of `int` in OP2Memory

DanRStevens:
Small fix I just happened to notice.


DanRStevens:
Seems there is more similar work to do. I opened issue #257 to track the changes. This is now part of the larger set of work.

PR #258, opened 2019-11-01
Op2Memory `size_t` updates

### LINKED ISSUE #257: Use `std::size_t` throughout OP2Memory
The OP2Memory code originally used `void*` to represent arbitrarily typed memory addresses. This led to a lot of casting of hardcoded memory addresses to `void*`. Instead, the library could take `std::size_t`, which would not require casting for hardcoded values. As the majority of values are hardcoded values determined from assembly analysis, not having to cast them would make the library code easier to work with. It should also reduce the need for internal type casting.

DanRStevens:
This closes #257.

### DIFF
--- a/srcDLL/AppEvents.cpp
+++ b/srcDLL/AppEvents.cpp
@@ -28,7 +28,7 @@ public:
 // By default, for plain functions, the caller cleans the stack, rather than the callee
 HINSTANCE WINAPI LoadShell(LPCSTR lpLibFileName);
 
-const auto loadLibraryDataAddr = reinterpret_cast<DWORD*>(0x00486E0A);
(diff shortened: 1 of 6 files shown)


DanRStevens:
The values represent memory addresses. Early on, I decided to use pointer syntax because of that. As the exact type of data was unknown and often didn't even matter, I went with `void*`. I didn't really know about `std::size_t` at the time, or at least didn't like it because I thought the name was awful. However, a `std::size_t` is defined to be able to span the memory space. Hence it is effectively able to hold memory addresses. It is also untyped, in the sense that it represents a memory address as a numerical value, without regard for the type at that address. It works directly on the address, not on what is being addressed. This turns out to be a much closer fit to how the code works. It manipulates addresses more so that what is at those addresses.

PR #264, opened 2019-11-03
Add missing header includes for `std::size_t`

DanRStevens:
I noticed a few places making use of `std::size_t` that don't seem to have a suitable header include for it. In all cases, the needed definition was pulled in indirectly through other header includes, hence no compile errors. However, such inclusions are subject to change without notice, so best to add a direct include.

For this, I considered indirect inclusion through an associated header file to be close enough:
SomeFile.cpp -> SomeFile.h -> cstddef
SomeFile.test.cpp -> SomeFile.h -> cstddef

All other files with more indirect includes were updated.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -14,6 +14,7 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <string>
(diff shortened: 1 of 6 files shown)


DanRStevens:
As have I.

PR #265, opened 2019-11-03
`constexpr` updates

DanRStevens:
Most patches will be written with hardcoded `constexpr` values, so I went and marked them as such.

A few constants were moved into the methods that used them. New jump table entries that must remain long-lived at fixed addresses (such as global/static variables) were marked as such.

### DIFF
--- a/srcDLL/DllMain.cpp
+++ b/srcDLL/DllMain.cpp
@@ -90,11 +90,12 @@ BOOL WINAPI DllMain(HINSTANCE hInstance, DWORD dwReason, LPVOID reserved)
 bool InstallDepPatch()
 {
 	// Set the execute flag on the DSEG section so DEP doesn't terminate the game
-	constexpr std::size_t destinationBaseAddress = 0x00585000;
-	bool success = Op2UnprotectMemory(destinationBaseAddress, 0x00587000 - 0x00585000);
+	constexpr std::size_t dsegSectionBaseAddress = 0x00585000;
+	constexpr std::size_t dsegSectionSize = 0x00587000 - 0x00585000;
+	bool success = Op2UnprotectMemory(dsegSectionBaseAddress, dsegSectionSize);
 
 	if (!success) {
-		LogError("Error unprotecting memory at: 0x" + AddrToHexString(destinationBaseAddress));
+		LogError("Error unprotecting DSEG memory at: 0x" + AddrToHexString(dsegSectionBaseAddress));
 	}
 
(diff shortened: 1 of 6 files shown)


DanRStevens:
In practice, there is often little to no difference between `constexpr` and `const`. If you declare a variable as `constexpr` it is also `const`. There may be runtime initialization or compile time initialization for either. In particular, for local variables on the stack, they may only be runtime initialized since stack space for the variable won't be allocated until the function runs. However runtime initialization is only done if space for the local variable is actually allocated. For `const` (and `constexpr`) values, if you never take the address of the variable, the compiler is free to inline all uses of it and not allocate space for the variable. The main difference is that a `const` variable's value may be runtime calculated (and then used to initialize the variable). On the other hand, a `constexpr` will give a compile time error if calculation of the value at compile time is not possible (even if runtime initialization is still used). For `constexpr` values, they may be used in certain special contexts, such as array size declarations, or template parameters. You may also use a `const` value in such a context as well, provided that it is initialized in a `constexpr` manner (even if it is not declared as such).

Example:
```cpp
// The right-hand-side may be reduced to a single constant value at compile time
// The actual assignment may be deferred until runtime for a local variable
constexpr auto someVariable = (someComplicatedExpressionReducedToSingleValue);
```

Example:
```cpp
// Function must be constexpr
constexpr auto compileTimeInitialized = CallSomeFunction();

// Likely runtime initialized, even if function is constexpr
const auto runTimeInitialized = CallSomeFunction();
```

----

For the code in question here, it really could have been done either way. I don't believe there is any compiler optimization that `constexpr` would enable here that `const` would not also allow. With that said, it felt semantically more meaningful to use `constexpr` here. The addresses being used were all set when the main executable was last compiled, and are all known at the time the DLL code is being compiled.

As a minor point, it also makes those addresses slightly easier to `grep` for. There are more contexts that make use of `const` than `constexpr`, such as `const` functions, and `const` parameters to functions. You could also try searching for `0x004` instead, but that misses the DEP patch addresses which start with `0x005`. If you instead search for `0x00`, you get extra spurious results from byte array initialization code. Being able to search for `constexpr` is easier.

In terms of moving forward, I'm fine with things just being declared as `const` unless there is a specific reason to use `constexpr`. I'm also fine with using `constexpr` where `const` would have been sufficient, if you feel it's a better semantic match. The compiler will tell you if the choice doesn't work. My preference is to use `constexpr` when assigning hardcoded address constants.

Issue #195, opened 2019-10-07
Use `unique_ptr` for Windows `HANDLE`


DanRStevens:
Similar to what was done in `LocalResource.h`, we could implement a `unique_ptr` wrapper for Windows `HANDLE` values. This could ensure automatic cleanup, and allow us to use exceptions a little more freely with code that uses Windows `HANDLE` values.

Although the `HANDLE` type is a single type seen by the compiler, there are conceptually a few different types of `HANDLE` values. In particular, some Windows API methods return `INVALID_HANDLE_VALUE` (-1) when an error occurs. Other methods will return `NULL` (0). Effectively, these are two types of `HANDLE` that have a different underlying value for what is effectively the `nullptr`. For both of them, we may need to call `CloseHandle`, though the check for when we have a valid resource may be different between them. As methods that use different conventions are not likely to intermix with each other, we don't need to worry about multiple `nullptr` like values simultaneously. Indeed, it would be bad practice to re-use the same `HANDLE` type variable with methods following a different convention.

I suggest using different wrapper types to distinguish the different `nullptr` value cases. The use of a type will depend on the documentation of the Windows API method being used, and in particular, the return value section of the documentation. To encode this type, the `nullptr` value can become a template parameter. Example:
```cpp
template <HANDLE nullValue>
struct HandleDeleter {
        using pointer = HANDLE;

        void operator()(HANDLE handle) const {
                if (handle != nullValue) {
                        CloseHandle(handle);
                }
        }
};

using UniqueHandleOrNull = std::unique_ptr<HANDLE, HandleDeleter<NULL>>;
using UniqueHandleOrInvalid = std::unique_ptr<HANDLE, HandleDeleter<INVALID_HANDLE_VALUE>>;
```

----

There are other API methods that use similar Windows types akin to `HANDLE`. In particular, `LoadLibrary`, which returns an `HMODULE`, which is a `typedef` for an `HINSTANCE`, which is a `typedef` for a `HANDLE`. In this case though, the `Deleter` would need to call `FreeLibrary` rather than `CloseHandle`. Perhaps this work can be undertaken together, due to the similarity.

----

We might also consider deriving from `unique_ptr` as a base class, rather than just a `using` alias. That would allow creating new methods, such as `empty()` to test if a pointer contains the null value for the given type.

DanRStevens:
[...]

PR #266, opened 2019-11-05
Add WindowsUniqueHandle.h

DanRStevens:
This completes the core of issue #195, creating an RAII style class for Windows `HANDLE` values.

### DIFF
--- a/srcStatic/WindowsModule.cpp
+++ b/srcStatic/WindowsModule.cpp
@@ -8,6 +8,8 @@
 #undef _UNICODE
 #include <windows.h>
 #include <tlhelp32.h> // CreateToolhelp32Snapshot, Module32First, Module32Next
+#include "WindowsUniqueHandle.h"
+
 
 std::string FindModuleName(HANDLE hModuleSnap, const void* address);
 bool containsAddress(MODULEENTRY32 const& moduleEntry, const void* address);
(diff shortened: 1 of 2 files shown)


srcStatic/WindowsModule.cpp
```diff
@@ -8,19 +8,20 @@
...
 	// Get all modules for current process (processId can be 0)
-	auto hModuleSnap = CreateToolhelp32Snapshot(TH32CS_SNAPMODULE, 0);
-	if (hModuleSnap == INVALID_HANDLE_VALUE) {
+	UniqueHandleOrInvalid hModuleSnap{CreateToolhelp32Snapshot(TH32CS_SNAPMODULE, 0)};
+	if (!hModuleSnap) {
 		return std::string("<Unable to create module snapshot>");
 	}
-	auto moduleName = FindModuleName(hModuleSnap, address);
-	CloseHandle(hModuleSnap);
+	auto moduleName = FindModuleName(hModuleSnap.get(), address);
```
Brett208: Might make sense to remove this variable and just directly return the result of FindModuleName.

```C++
return FindModuleName(hModuleSnap.get(), address);
```

srcStatic/WindowsUniqueHandle.h
```diff
@@ -0,0 +1,44 @@
...
+
+	explicit operator bool() {
+		return (reinterpret_cast<HANDLE>(nullValue) != base::get());
+	}
+};
+
+
+// For APIs that return INVALID_HANDLE_VALUE (-1) on error
+// Replace INVALID_HANDLE_VALUE macro with an equivalent constexpr value
+using UniqueHandleOrInvalid = UniqueHandle<-1>;
```
Brett208: What do you think about the name UniqueOrInvalidHandle? I think it is easier to read in this form.

DanRStevens: Err, not really, no. The "Unique" part isn't really part of the "Or". It's more like "Unique(Handle | Invalid)".

srcStatic/WindowsModule.cpp
DanRStevens: Good point. I wasn't even looking at the code at that high of a level.

PR #268, opened 2019-11-05
Change pointer integer types to `uintptr_t`

DanRStevens:
I recently came across a few sources that point to the subtle distinction between `std::size_t` and `uintptr_t`. It seems some of our uses of `std::size_t` are perhaps more accurately handled by the `uintptr_t` type.

Technically pointers to objects and maximum size of objects can be two different types with two different sizes. It just so happens that in today's typical flat memory models, the two types coincide. Previously, on segmented memory models, this was not the case. For 16-bit x86 real mode programming, object size might be restricted to a 16-bit segment, while an address might include both a segment and an offset, resulting in a 32-bit pointer. Future architectures are similarly not restricted to have matching types/sizes.

Reference:
[Fixed width integer types](https://en.cppreference.com/w/c/types/integer)  (C99, included by C++11)
[What is uintptr_t data type](https://stackoverflow.com/questions/1845482/what-is-uintptr-t-data-type)

There are likely other instances throughout the project that could be updated.

### DIFF
--- a/srcStatic/OP2Memory.cpp
+++ b/srcStatic/OP2Memory.cpp
@@ -6,8 +6,8 @@
 
 
 bool memoryPatchingEnabled = false;
-uintptr_t loadOffset = 0;
-constexpr uintptr_t ExpectedOutpost2Addr = 0x00400000;
+std::uintptr_t loadOffset = 0;
+constexpr std::uintptr_t ExpectedOutpost2Addr = 0x00400000;
 
 
 // Enabled patching of Outpost2.exe memory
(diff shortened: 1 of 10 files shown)


srcStatic/OP2Memory.cpp
```diff
@@ -6,8 +6,8 @@
 
 
 bool memoryPatchingEnabled = false;
-std::size_t loadOffset = 0;
-constexpr std::size_t ExpectedOutpost2Addr = 0x00400000;
+uintptr_t loadOffset = 0;
```
Brett208: What do you think about labeling all these std::uintptr_t?

I also noticed we have some uint64_t type definitions in OP2Utility. I was going to ask if they should be prepended to read std::uint64_t.

DanRStevens: Good question. All the examples I saw were without a `std::` prefix. That seems to stem from these coming from C99. The C language doesn't have namespaces, only C++ does.

Turns out the `std::` prefix seems to work.

I suppose I find it a bit of a nuisance when seemingly primitive data types must be namespace prefixed, though that is the suggestion for `std::size_t`. I may follow up with an update once I've convinced myself it really is standard, and proper to do.

Brett208: I had already tested the std namespace on the types; it works on MSVC as well. 

It doesn't bother me that we are not using it, besides that it seems like a double standard when we do it for size_t but not others.

Compiled and worked fine on my local machine.

DanRStevens:
I expanded a bit on the use of `uintptr_t`. I also did a bit of reading where I found many examples of `std::uintptr_t`. That does indeed seem like the correct C++ way of doing things.

DanRStevens:
Nah, you're right. I think the `std::` prefix is right for C++. I just happened to only come across C examples when I first found the type.

I suppose it would have been nice if those primitive types could have been added to the language at the same time as `int`, `short`, and `char`. (They maybe they wouldn't need the `_t` suffix either). Oh well, what can you do. There's a good potential it could have broken existing code if they have added new names into the global scope. Preserving the ability to compile old code seems to have been a priority for the C++ extensions.

PR #272, opened 2019-11-07
Add `UniqueModule`

### LINKED ISSUE #195: Use `unique_ptr` for Windows `HANDLE`

DanRStevens:
This closes #195. This completes the secondary minor point in that issue about a corresponding `std::unique_ptr` based RAII style class to handle `HINSTANCE` values returned by `LoadLibrary`.

### DIFF
--- a/srcStatic/WindowsUniqueModule.h
+++ b/srcStatic/WindowsUniqueModule.h
@@ -1,0 +1,18 @@
+#include <windows.h>
+#include <memory>
+
+
+// Sentinel value guarded HMODULE deleter
+// For use with LoadLibrary (will automatically call FreeLibrary)
+struct ModuleDeleter {
+	using pointer = HMODULE;
+
+	void operator()(HMODULE handle) const {
+		if (handle != NULL) {
+			FreeLibrary(handle);
+		}
+	}
+};
+
+
+using UniqueModule = std::unique_ptr<HMODULE, ModuleDeleter>;
--- a/srcStatic/GameModules/DllModule.cpp
+++ b/srcStatic/GameModules/DllModule.cpp
@@ -10,16 +10,14 @@ DllModule::DllModule(const std::string& moduleName)
 void DllModule::LoadModuleDll(const std::string& dllPath)
 {
 	// Try to load a DLL with the given name (possibly an empty string)
-	HINSTANCE dllHandle = LoadLibraryA(dllPath.c_str());
+	moduleDllHandle.reset(LoadLibraryA(dllPath.c_str()));
 
-	if (!dllHandle) {
+	if (!moduleDllHandle) {
 		throw std::runtime_error(
 			"Unable to load DLL path: " + dllPath +
 			" LoadLibrary " + GetLastErrorString()
 		);
 	}
-
-	this->moduleDllHandle.reset(dllHandle);
 
 	DetectExportedModuleFunctions();
 }
--- a/srcStatic/GameModules/DllModule.cpp
+++ b/srcStatic/GameModules/DllModule.cpp
@@ -19,7 +19,7 @@ void DllModule::LoadModuleDll(const std::string& dllPath)
 		);
 	}
 
-	this->moduleDllHandle = dllHandle;
+	this->moduleDllHandle.reset(dllHandle);
 
 	DetectExportedModuleFunctions();
 }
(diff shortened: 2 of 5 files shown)


srcStatic/WindowsUniqueModule.h
```diff
@@ -0,0 +1,18 @@
...
+#include <memory>
+
+
+// Sentinel value guarded HMODULE deleter
+// For use with LoadLibrary (will automatically call FreeLibrary)
+struct ModuleDeleter {
+	using pointer = HMODULE;
+
+	void operator()(HMODULE handle) const {
+		if (handle != NULL) {
```
Brett208: Should this be nullptr instead of NULL. Or maybe it could be 

```C++
if (handle) {
    ...
}
```

DanRStevens: Hmm, interesting point. I do like that simplification. I suppose technically correctness depends on the definition of `NULL`, which we aren't in control of. Though realistically, the definition of `NULL` isn't about to change.

I checked real quick what the zero-as-null-pointer warning flag has to say about it. No warning is produced.

I think I'm tempted to leave this given the definition of `NULL` is external.

Brett208: Could you expand on your thought process? I'm not following why we would not use the keyword definition of nullptr to check if a handle is null (I think of a handle as a pointer).

-Brett

DanRStevens: A `HANDLE` is currently implemented as a pointer, though it could also be implemented as a numeric integer value. It's an opaque type. You don't really know how it's implemented, and the internal details could change. They probably won't, but conceivably a future (or past) version of Windows could choose to use a different internal representation. They would also be free to change the sentinel `NULL` value. Instead of it being `0` or `nullptr`, it could become some other magic number, such as `-1`. If the definition of `NULL` changed, and we weren't using `NULL` itself, the code would no longer be correct.

Brett208: Interesting, that article talks about wrapping handles like you are doing here.

DanRStevens:
Ok, I'm going to merge this now.

Tagging issue #275 as somewhat related to the discussion that was brought up here.

PR #276, opened 2019-11-11
Update `GetConsoleModDir_s` to use `FindModuleHandle`

### LINKED ISSUE #186: Update `GetConsoleModDir_s` specification
If we allow multiple console module, there is no longer a clear single console module directory. As such, we need to update `GetConsoleModDir_s` to account for this.

One possible way to update this, is to use the same `_ReturnAddress()` trick from the `Log` method to lookup which module the call came from, and return the directory for that module. That allows a console module to lookup its own folder.

In case the call comes from a non-console module, or even a non-module, we can perhaps return either `""`, or return the folder of the first loaded console module, if any. The former is perhaps a bit more sensible, though the later would avoid certain breaks with traditional behaviour. I'm not sure such traditional behaviour was ever used, or was ever useful though.

DanRStevens:
This closes #186.

### DIFF
--- a/srcStatic/WindowsModule.h
+++ b/srcStatic/WindowsModule.h
@@ -2,6 +2,7 @@
 
 #include <string>
 #include <intrin.h> // _ReturnAddress
+#include <windows.h>
 
 
 #pragma intrinsic(_ReturnAddress)
@@ -11,3 +12,4 @@
 
 
 std::string FindModuleName(const void* address);
+HMODULE FindModuleHandle(const void* address);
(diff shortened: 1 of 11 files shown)


srcStatic/WindowsModule.h
```diff
@@ -10,5 +11,5 @@
 #endif
 
 
-// Finds the module which contains the given address
 std::string FindModuleName(const void* address);
+HMODULE FindModuleHandle(const void* address);
```
Brett208: Should this return a WindowsUniqueModule instead of an HModule?

DanRStevens: No, because that would denote a transfer of ownership.

Issue #250, opened 2019-10-27
Create a subdirectory for Log related files


Brett208:
We have about 11 sourcecode files associated with logging code in op2extStatic. It would be nice to push most or all of those into a subdirectory to improve project organization.

For MSVC, I could followup the change with adding a new filter for the IDE so they show in a folder within Visual Studio.

DanRStevens:
This would be good to do once PRs #251 and #252 are closed off.

DanRStevens:
This might be good to do now. There are currently no open pull requests that modify logging code.

I was thinking all the `Logger` classes could be moved into a "Log/" subfolder. Perhaps the Log.h and Log.cpp files can be left at top level. The Log.h file is the main include for files that need logging support, so it may be nice to keep it at the top level. There is also an implementation and visibility difference between the Log.h/.cpp files and the Logger implementation classes. The main Log.h file is needed by logging clients. The Logger.h file is needed by logger implementations.

```
Log.cpp
Log.h
Log/
  LoggerDebug.cpp
  LoggerDebug.h
  LoggerDistributor.cpp
  LoggerDistributor.h
  LoggerFile.cpp
  LoggerFile.h
  Logger.h
  LoggerMessageBox.cpp
  LoggerMessageBox.h
```

We could rename the `Logger` classes to just `Log`. After updating filenames to match, we would have two "Log.h" files, however, they would be in different folders. I think that would be acceptable, though might be considered a bit ambiguous.

Brett208:
I'm not sure I like the idea of two files named Log.h. I don't have time to review and make a suggestion right now though. Maybe tomorrow I can look at it.

Brett208:
How about LogInterface.h as a replacement name for Logger.h? That would keep from having duplicate filenames within the project. I think LogInterface.h is a better name than Logger.h.

I would slightly prefer seeing Log.h/.cpp within the log directory. However, I am okay with the solution presented here.

-Brett

DanRStevens:
Hmm, what if we named the directory a bit more precisely, such as `LogDestinations/`? Or maybe `LogInterfaces/`? Though that might be a bit of a misnomer if it mostly contains concrete derived classes.

I think I would prefer keeping "Interface" out of class and filenames.

Maybe there is a better name for the original Log.h/Log.cpp?

Brett208:
Naming the directory Log works for me. Although I'm okay with whatever you are thinking for the directory name.

DanRStevens:
[...]

PR #277, opened 2019-11-11
Create "Log/" subfolder

DanRStevens:
This addresses the move of the `Logger` classes in to `Log/" subfolder from issue #250. It does not address renaming any of the classes.


DanRStevens:
Ahh, thank you! I was going to ask for help with that, but it seems I forgot.

PR #283, opened 2019-11-18
Misc fixes

DanRStevens:
Fixed a few things reported by increased warning options from Mingw.

### DIFF
--- a/makefile
+++ b/makefile
@@ -9,7 +9,7 @@ include makefile-generic.mk
 config := mingw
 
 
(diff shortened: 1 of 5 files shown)


DanRStevens:
There does need to be a semicolon after the closing brace of a class declaration. Same with structs and unions too. The code here though is the closing brace of the constructor and class method implementation in the .cpp file.

DanRStevens:
I added one more flag for warning checks to the Linux build. Now that the debug logger change to always log for all configurations has been merged into master, that code will no longer trigger a warning about an unused parameter for non-debug builds.

The flag change is a Linux only change.

PR #284, opened 2019-11-18
Remove conditional compile check for debug logging

DanRStevens:
Logging to an attached debugger should be allowed under any build configuration.

This is something I wanted to change for ages, but wanted to keep it separate from other changes.


DanRStevens:
Yeah, the code is barely outputting anything. As for where it goes, there's a hook point for an attached debugger. This is how OllyDbg or Visual Studio are able to capture those messages. If no debugger is attached, the messages are discarded by the OS.

PR #286, opened 2019-11-26
Add AppVeyor build artifacts and release tagging

DanRStevens:


Brett208:
Before merging this, I'd like to have a discussion on planned final objectives. My current thoughts are:

1. Prevent the debug artificats from being copied into a release. (Fine if they are present within appveyor for experimentation, just not copied to the formal release page). By debug I mean the debug configuration build, not the PDB file for the release compilation, which I think should stay for future troubleshooting on the release page (basic symbols library)..

2. I think we are in agreement to attempt to line up the zip filename to use the tagged release instead of the appveyor release number, although unsure of the exact syntax/strategy to accomplish.

3. I was thinking about bundling or not bundling the readme more. After some more thought, I think we should consider not including it. I see plenty of projects that do not bundle the readme with the binary distribution. We could just add a reference or link to the repository readme in the release page documentation and forgo including it with the binary. This would:
    1. Simplify the post build script slightly.
    1. Allow including things like pictures and build status monitoring graphics (if we ever decide to) to the readme without worrying about those assets being absent from the binary readme.md package.
    1. Keep the end user from choosing an arbitrary application to open a markdown syntax document on Windows. (Currently, there isn't a default program, so the user would have to guess on notepad, or a web browser, etc, which for a non-technically savvy person would be confusing, although certainly achievable). They could just be directed to the GitHub page and the readme would be displayed properly with markdown.

After deciding what to do here, it would be nice to consider translating to MissionScanner to handle a dual x86/x64 release.

Open to comments and changes to above thoughts

Brett208:
Before merging this, I'd like to have a discussion on planned final objectives. My current thoughts are:

1. Prevent the debug artificats from being copied into a release. (Fine if they are present within appveyor for experimentation, just not copied to the formal release page). By debug I mean the debug configuration build, not the PDB file for the release compilation, which I think should stay for future troubleshooting on the release page (basic symbols library)..

2. I think we are in agreement to attempt to line up the zip filename to use the tagged release instead of the appveyor release number, although unsure of the exact syntax/strategy to accomplish.

3. I was thinking about bundling or not bundling the readme more. After some more thought, I think we should consider not including it. I see plenty of projects that do not bundle the readme with the binary distribution. We could just add a reference or link to the repository readme in the release page documentation and forgo including it with the binary. This would:
    1. Simplify the post build script slightly.
    1. Allow including things like pictures and build status monitoring graphics (if we ever decide to) to the readme without worrying about those assets being absent from the binary readme.md package.
    1. Keep the end user from choosing an arbitrary application to open a markdown syntax document on Windows. (Currently, there isn't a default program, so the user would have to guess on notepad, or a web browser, etc, which for a non-technically savvy person would be confusing, although certainly achievable). They could just be directed to the GitHub page and the readme would be displayed properly with markdown.

After deciding what to do here, it would be nice to consider translating to MissionScanner to handle a dual x86/x64 release.

Open to comments and changes to above thoughts

DanRStevens:
I removed your duplicate comment.

I think these updates address the concerns you presented.

DanRStevens:
You're welcome to test it out. I did some testing in the other repository before bringing the code over (and deleted the releases after I was done). Barring any errors in porting the changes over, it should work.

----

Maybe at some point we can add auto detection of .lib outputs, perhaps through the absence of .exe or .dll outputs. I'm not looking to address that at this time though, and it can't really be tested as part of this repository anyway.

PR #287, opened 2019-11-29
Store Linux artifacts

DanRStevens:
Store Linux artifacts on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,5 +13,6 @@ jobs:
       - run: make --keep-going --jobs=2 testDll
       - run: make --keep-going --jobs=2 check-testLib
       - run: make --keep-going --jobs=2 check-testDll
+      - run: tar -czf op2ext.tar.gz op2ext.dll
       - store_artifacts:
-          path: op2ext.dll
+          path: op2ext.tar.gz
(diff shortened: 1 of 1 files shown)


Brett208:
@DanRStevens, I cannot seem to find the artifacts for this build. I think they are supposed to be in a tab next to the configuration tab, but I don't see it.

Also, how does this work on Linux sense there are multiple available operating systems (Debian, CentOS, Fidora, etc)? Is the generated artifact only acceptable for use on a subset of Linux systems, or is it okay with any modern operating system? 

-Brett

DanRStevens:
CircleCI is considerably less obvious about artifacts. When you go to the artifacts tab, it expands some of the folders into a tree view, however the last folder level appears to be collapsed, and so it hides the file attachment. The arrows indicating which levels are expanded or collapsed are not totally obvious.

I'm afraid I'm a little uncertain about how executables work with various distributions of Linux. The following is a bit of a guess. The distributions have differences in expected libraries, and perhaps even install locations. That could easily impact dynamic linking. However, I believe this project was built with static linking, so none of that should matter. I think the main requirement is a minimum Kernel version, and given that we don't use any particularly special or advanced feature, I assume the minimum would be pretty old and widespread by now.

DanRStevens:
Ahh, thanks for the info. At least we know how this stuff works now.

Oh, and technically, the Linux build cross compiles for Windows from Linux, so you could potentially use the Mingw produced DLL from Windows. On Linux, it's all essentially Windows code being run through Wine. I hadn't really thought about that before.

PR #289, opened 2019-12-22
Allow calling LogError and LogDebug from external DLLs

### LINKED ISSUE #64: Allow for verbose logging
What sort of strategy do we want for allowing verbose/diagnostic logging? We can either support different logging details at the per module level or overall for all modules and op2ext.

We could add a property to the .ini file

```
[Game]
Logging=Verbose
```

I think for now a Default and Verbose option would be sufficient.

We could then add a function to op2ext.h:

```C++
enum LoggingLevel {
    Default,
    Verbose
}

LoggingLevel GetLoggingLevel()
```

The module would call GetLoggingLevel on initialization and record the value. It would then be up to the module to determine when a message met the fetched requirement.

---

Another solution would be for each module to check its section of the .ini file for logging verbosity. This would give more fine grained control (one module could be verbose, and one default). However Console modules would not be able to support verbosity through the .ini file since they do not use it. They could pull a global setting through op2ext's public interface though.

Brett208:
Closes #64 

I wanted input on the external function names before writing test code.

The Issue suggested:
```C++
Log(char* message); // Normal logging
LogError(char* message); // Very serious error, maybe with a popup
LogDebug(char* message); // For really verbose stuff that may not be so interesting
```

Which I would be fine with. However the private template function I wrote throws an error because it cannot distinguish between two overloads for LogError. One version being the internal function within op2ext and the other being now contained in op2ext.h for other dlls to call.

### DIFF
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -120,20 +120,17 @@ void GenericLog(const char* message, const void* moduleAddress, LogFunctionType 
 
 OP2EXT_API void Log(const char* message)
 {
-	const void* moduleAddress = _ReturnAddress();
-	GenericLog(message, moduleAddress, LogMessage);
+	GenericLog(message, _ReturnAddress(), LogMessage);
 }
 
 OP2EXT_API void LogError(const char* message)
 {
-	const void* moduleAddress = _ReturnAddress();
-	GenericLog(message, moduleAddress, LogError);
+	GenericLog(message, _ReturnAddress(), LogError);
 }
 
 OP2EXT_API void LogDebug(const char* message)
 {
-	const void* moduleAddress = _ReturnAddress();
-	GenericLog(message, moduleAddress, LogDebug);
+	GenericLog(message, _ReturnAddress(), LogDebug);
 }
 
 
--- a/srcDLL/op2ext.cpp
+++ b/srcDLL/op2ext.cpp
@@ -91,7 +91,9 @@ OP2EXT_API void SetSerialNumber(char major, char minor, char patch)
 	}
 }
 
-OP2EXT_API void Log(const char* message)
+
+template <typename LogFunction>
+void GenericLog(const char* message, LogFunction logFunction)
 {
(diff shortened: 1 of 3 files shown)


Brett208:
Thanks for the feedback. You were correct about the address displaying incorrect (everything was showing as from op2ext). 

Now should be able to write test cases.

srcDLL/op2ext.cpp
```diff
@@ -120,17 +120,20 @@ void GenericLog(const char* message, LogFunction logFunction)
 
 OP2EXT_API void Log(const char* message)
 {
-	GenericLog(message, LogMessage);
+	const void* moduleAddress = _ReturnAddress();
+	GenericLog(message, moduleAddress, LogMessage);
```
DanRStevens: Should be fine to combine these two lines and get rid of the `moduleAddress` temporary.

Actually, the name `moduleAddress` is even slightly wrong, since the address is not the base address of the module, but rather the address of the instruction after the `CALL` instruction that called into this function. That will be somewhere within the module's address space, but certainly not at the base address.

Same for the other call points.

Brett208: fixed by removing the temporary variable. I wasn't sure if mingw would be okay with calling the temporary as an argument on a function.

Brett208:
@DanRStevens, I know you approved this, but I had to comment out testing the LogError test call as it will produce a modal dialog. It works fine on my CPU, but would probably break automated unit test calls on the build servers. I'm good to merge if this is okay with you.

DanRStevens:
Err, are you sure about that? The test project shouldn't generate message boxes.

Brett208:
I was trying to test as you would use a proper module (which I think means the module should not be manipulating the static project). Can we call the static project functions, affect op2ext, and still consider it a valid DLL cross boundary test?

-Brett

DanRStevens:
Yeah, I was just thinking that does get a bit messy to test. Ehh, this is probably fine for now. I don't really want to muck with it.

PR #292, opened 2019-12-24
Rename logger

### LINKED ISSUE #231: Better distinguish between Log and Logger classes
I find it difficult to distinguish between a Log and Logger in the code base.

Maybe we could rename Logger to ILog or LogInterface? ILog is pretty standard in C# conventions, but LogInterface might be better here.

This would look like:
```C+++
void SetLogInterface(LogInterface* newLogInterface);
void SetErrorLogInterface(LogInterface* newLogInterface);
```

I would prefer SetErrorLogInterface to SetLogErrorInterface, however, I could live with either option.

Open to other suggestions as well. I've tried to adapt to using the current syntax (Log and Logger), but it doesn't seem to be sticking for me.

Brett208:
Closes #231 

I think this is an improvement overall on readability, although there may be room to tweak some of the names I chose..


srcStatic/op2extStatic.vcxproj
```diff
@@ -167,7 +167,7 @@
     <ClInclude Include="IniFile.h" />
     <ClInclude Include="LocalResource.h" />
     <ClInclude Include="Log.h" />
-    <ClInclude Include="Log/Logger.h" />
+    <ClInclude Include="Log\LogDestination.h" />
```
DanRStevens: What's up with the path separator on this line? It doesn't match the others.

srcStatic/op2extStatic.vcxproj.filters
```diff
@@ -90,6 +87,9 @@
     <ClInclude Include="GameModules\EarthworkerProximityTasking.h">
       <Filter>GameModules</Filter>
     </ClInclude>
+    <ClInclude Include="Log\LogDestination.h">
```
DanRStevens: The path separator changed here too. Though there are other similar entries with the same slash.

srcStatic/op2extStatic.vcxproj
```diff
@@ -140,10 +140,10 @@
...
     <ClCompile Include="IniFile.cpp" />
     <ClCompile Include="Log.cpp" />
-    <ClCompile Include="Log/LoggerDebug.cpp" />
-    <ClCompile Include="Log/LoggerDistributor.cpp" />
-    <ClCompile Include="Log/LoggerFile.cpp" />
-    <ClCompile Include="Log/LoggerMessageBox.cpp" />
+    <ClCompile Include="Log\LogDebug.cpp" />
+    <ClCompile Include="Log\LogDistributor.cpp" />
+    <ClCompile Include="Log\LogFile.cpp" />
+    <ClCompile Include="Log\LogMessageBox.cpp" />
```
DanRStevens: Path separator changed here too. This seems to make things consistent again.

srcStatic/op2extStatic.vcxproj
```diff
@@ -168,10 +168,10 @@
...
     <ClInclude Include="Log.h" />
     <ClInclude Include="Log\LogDestination.h" />
-    <ClInclude Include="Log/LoggerDebug.h" />
-    <ClInclude Include="Log/LoggerDistributor.h" />
-    <ClInclude Include="Log/LoggerFile.h" />
-    <ClInclude Include="Log/LoggerMessageBox.h" />
+    <ClInclude Include="Log\LogDebug.h" />
+    <ClInclude Include="Log\LogDistributor.h" />
+    <ClInclude Include="Log\LogFile.h" />
+    <ClInclude Include="Log\LogMessageBox.h" />
```
DanRStevens: Same.

srcDLL/DllMain.cpp
```diff
@@ -29,10 +29,10 @@ void OnShutdown();
...
 // Similarly these should not use globals from other files before DllMain has started
 // Pay careful attention to anything passed to a constructor, or called by a constructor
-LoggerFile loggerFile; // Logging to file in Outpost 2 folder
-LoggerMessageBox loggerMessageBox; // Logging to pop-up message box
-LoggerDistributor loggerDistributor({&loggerFile, &loggerMessageBox});
-LoggerDebug loggerDebug;
+LogFile logFile; // Logging to file in Outpost 2 folder
+LogMessageBox logMessageBox; // Logging to pop-up message box
+LogDistributor logDistributor({&logFile, &logMessageBox});
+LogDestinationDebug logDestinationDebug;
```
DanRStevens: Why does this one have "Destination" in the name? It seems inconsistent. Accidental search/replace?

srcStatic/Log.cpp
```diff
@@ -11,55 +11,55 @@ namespace {
...
+	LogDestination* logDestinationMessage = nullptr;
+	LogDestination* logDestinationDebug = nullptr;
 }
 
 
 // Set logging output destination for each logging level
-// Caller owns the logger and is responsible for cleanup when logger is no longer required
-// Use `SetLoggerX(nullptr);` to unset a logger
+// Caller owns the log destination and is responsible for cleanup when log destination is no longer required
+// Use `SetLogDestinationXXXX(nullptr);` to unset a log destination
```
DanRStevens: Something about all the Xs seems kind of funny here. Might be my math background talking.

srcDLL/DllMain.cpp
```diff
@@ -42,9 +42,9 @@ BOOL WINAPI DllMain(HINSTANCE hInstance, DWORD dwReason, LPVOID /*reserved*/)
 	// This will be called once the program is unpacked and running
 	if (dwReason == DLL_PROCESS_ATTACH) {
 		// Setup logging
-		SetLoggerError(&loggerDistributor);
-		SetLoggerMessage(&loggerFile);
-		SetLoggerDebug(&loggerDebug);
+		SetLogDestinationError(&logDistributor);
+		SetLogDestinationMessage(&logFile);
+		SetLogDestinationDebug(&logDestinationDebug);
```
DanRStevens: I'm not a huge fan of this rename, though maybe I just need to warm up to it. It's more precise to include "Destination", though it feels kind of long and wordy.

srcStatic/Log.cpp
```diff
@@ -11,55 +11,55 @@ namespace {
...
 
-// Set logger for error logging level
-void SetLoggerError(LogDestination* newLogger) {
-	loggerError = newLogger;
+// Set log destination for error logging level
+void SetLogDestinationError(LogDestination* newLogLogDestination) {
```
DanRStevens: Typo: `newLogLog...`

srcStatic/Log/LogDebug.h
```diff
@@ -3,7 +3,7 @@
 #include "LogDestination.h"
 
 
-class LoggerDebug : public LogDestination
+class LogDestinationDebug : public LogDestination
```
DanRStevens: This rename looks like a search/replace error.

srcStatic/op2extStatic.vcxproj
Brett208: I believe Visual Studio will default to using the Windows specific path separator when automatically editing the filters file, although it will work with either path separator. I'd like to default to using the Windows specific "\" since the IDE seems to generate it when updating tracked files. I suspect all the "/" path separators were added by manual edits to the filters file instead of the automated edits, although I'm not really inclined to research at this time.

srcStatic/op2extStatic.vcxproj.filters
Brett208: See response above. Likely done automatically by the IDE when I used it for edits.

srcStatic/op2extStatic.vcxproj
DanRStevens: I wonder if an earlier version of Visual Studio used a different path separator. I imagine people editing on Windows would have added files through Visual Studio, and people editing from Linux would just copy/paste from what was already present.

At any rate, it should probably be using the Visual Studio default.

srcDLL/DllMain.cpp
Brett208: LogDebug already exists as a global level function. I wasn't sure we wanted global function and a class with the same signature. Maybe there is an alternate name like LogDebugMessage that would be better?

srcStatic/Log.cpp
Brett208: I liked the multiple XXXXs look better than the single X previously, although it is really a minor thing to me (also not directly related to the other changes. If you prefer the single X I'll switch it back.

srcDLL/DllMain.cpp
Brett208: I agree it is wordy but also more precise. I'm okay with either leaving or pulling the word Destination.

srcStatic/Log.cpp
Brett208: Thanks for catching

srcStatic/Log/LogDebug.h
Brett208: It was on purpose. Would you prefer a different name?

srcStatic/Log.cpp
DanRStevens: I suppose it's a bit like `YYYY-MM-DD`. Of course those are fixed width fields, and we have a variable length name at the end. That might actually be what was bothering me about it. When I saw a specific number of Xs, I assumed the variable portion would be that length, but it's not.

srcDLL/DllMain.cpp
DanRStevens: I kind of want to pull it here. I find when its there, it changes how I break up and parse the words, and they make less sense. It's like I'm putting a slight pause between "Log" and "Destination". Mentally grouping "Destination" with "Error", "Message", or "Debug" just doesn't flow well, and makes the name sound a bit confusing. If someone else had no experience with the API, they might be tempted to parse the names in the same way.

srcStatic/Log/LogDebug.h
DanRStevens: Hmm, I see now. That is unfortunate. We can't have a class with the same name as a function. Though if we rename either one of them to include "Destination", we've introduced an asymmetry to the naming in our API. I kind of would like to rename this, though we may need to be creative. What if we changed the "Debug" part? Is there something more descriptive we could use there? Maybe `LogAttachedDebugger` for the class name?

Brett208:
@DanRStevens would you prefer one of us take another pass on this and then merge or would you prefer just closing the branch unmerged and close issue #231. At this point, I'm okay either way, but it seems you were perhaps not as enthusiastic about the new names.

DanRStevens:
Wow, I'd really lost context on this one.

I remember there were some naming oddities introduced here, which I kind of wanted to avoid. I don't remember so clearly why we wanted to rename stuff. I wasn't overly attached to some of the old names, so I'm fine with renaming things, though it seems we need to be careful about collisions between global function names and class names.

I could go either way on revisiting versus closing.

PR #296, opened 2020-03-08
Insert newline character automatically when calling LogDebug

Brett208:
I'm reading that whenever std::endl is called, std::flush is called. I think maybe we want to move away from endl? Unfortunatley, there doesn't appear to be a dropin Environment.Newline like in C# in C++.

### DIFF
--- a/srcStatic/Log/LoggerDebug.cpp
+++ b/srcStatic/Log/LoggerDebug.cpp
@@ -2,15 +2,9 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
-#ifdef _WIN32
 const std::string newline("\r\n");
-#elif defined macintosh // OS 9
-const std::string newline("\r");
-#else
-const std::string newline("\n") // Mac OS X uses \n
-#endif
 
 void LoggerDebug::Log(const std::string& message)
 {
-	OutputDebugString(std::string(message + newline).c_str());
+	OutputDebugString((message + newline).c_str());
 }
--- a/srcStatic/Log/LoggerDebug.cpp
+++ b/srcStatic/Log/LoggerDebug.cpp
@@ -2,8 +2,15 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
+#ifdef _WIN32
+const std::string newline("\r\n");
+#elif defined macintosh // OS 9
+const std::string newline("\r");
+#else
+const std::string newline("\n") // Mac OS X uses \n
+#endif
 
 void LoggerDebug::Log(const std::string& message)
 {
-	OutputDebugString(message.c_str());
+	OutputDebugString(std::string(message + newline).c_str());
 }
--- a/srcStatic/Log/LoggerDebug.cpp
+++ b/srcStatic/Log/LoggerDebug.cpp
@@ -2,7 +2,7 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
-const std::string newline("\r\n");
+const std::string newline("\n");
 
 void LoggerDebug::Log(const std::string& message)
 {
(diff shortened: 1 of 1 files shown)


DanRStevens:
Hmm, I'm used to `"\n"` being expanded to the platform specific newline anyway. I'm not sure there is any need to get fancy about this. Though I admit I'm a little uncertain when the expansion happens. I assume `"\n"` is stored as a single byte in a string, but then expanded when output through standard streams. I'm not sure if `OutputDebugString` counts for that.

I should point out though that `OutputDebugString` is platform specific, so it would be quite appropriate to use platform specific newlines with that method. Actually, this whole project is highly platform specific. It will never run on other platforms outside of some kind of emulation or compatibility layer.

Case in point, right above the `#ifdef _WIN32` is:
```cpp
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
```

----

I have heard that `std::endl` will also flush a stream. In some cases though, streams are already set to flush on newline, so there may be no difference. I don't think it typically matters for `std::cout`, though might matter for a file stream. At any rate, we do so little logging, it's unlikely we would ever notice a performance difference. I would be more concerned with safety and style, unless there was a massive amount of output being done.

Brett208:
I don't think std::endl is compatible with std::string concatenation

Brett208:
I'm testing using the output window from Visual Studio 2019.

OutputWindow

Visual Studio Output will properly display line endings with either "\r\n" or just "\n". I figured since this is compiled for Windows we would use the Windows line ending. I'm uncertain why we wouldn't use the Windows version, but if it works, I'm not too concerned one way or the other.

I would like endings in the output for debugging and am not too concerned with the implementation at this point. I think @DanRStevens should take over this branch and implement as desired.

DanRStevens:
What happens without the line endings? A method like that might include implicit line endings for all strings.

I'm afraid this one would be a little difficult for me to test. I could probably get output through OllyDbg, though that's not nearly as convenient for development as the MSVC debug window.

srcStatic/Log/LoggerDebug.cpp
```diff
@@ -2,8 +2,9 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
+const std::string newline("\r\n");
 
 void LoggerDebug::Log(const std::string& message)
 {
-	OutputDebugString(message.c_str());
+	OutputDebugString((message + newline).c_str());
```
DanRStevens: Question, was this change to make output function *code* more symmetric? (Without regard to actual *displayed* output?) If that's the case, I would recommend no change.

I'm unclear on whether or not  `OutputDebugString` provides automatic line termination. It seems likely a function like that would cause messages to be automatically separated by lines. I suppose it might depend on the attached debugger though.

Given that we're having this discussion though, I assume actual output alerted you to the difference (as opposed to source code asymmetry), and that `OutputDebugString` does not result in automatic newlines. If that is the case, then I would suggest:
```suggestion
	OutputDebugString((message + "\n").c_str());
```

That should be quite understandable to the majority of developers, will little syntactic noise, and no cross referencing of definitions.

Brett208: OutputDebugString does not line terminate provided input. I'll push your suggestion momentarily.

srcStatic/Log/LoggerDebug.cpp
```diff
@@ -2,8 +2,7 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
-
```
DanRStevens: Pretty minor point, but we lost a blank line here.

Brett208:
See below if no newline character is added. I may redo this branch so there will be less commit noise onto master.

OutputNoNewline

DanRStevens:
Ok, looks like the newline is pretty necessary then.

PR #297, opened 2020-03-15
Add newline character to debug messages

Brett208:


DanRStevens:
Adding reference to earlier discussion and work: #296

PR #298, opened 2020-06-09
Update gtest nuget package

Brett208:


Brett208:
Thanks, I was planning to commit the changelog to this branch as well. I'm trying to figure out if we actually broke any public interfaces and need to call it 3.0.0 instead of 2.3.0.

Brett208:
OK, just found the breaking change.

PR #299, opened 2020-06-10
Improve documentation of GetConsoleModDir_s & GetConsoleModDir

Brett208:
To prevent breaking change, the function name has not changed, but now the function can pull the directory of ini modules.


Brett208:
Done for the day. Still working on the change log. I have the bulk of the change log written, but want to add examples on how to load multiple console modules and new format in .ini file before pushing.

Brett208:
Yeah, it was fixed somewhere between 2.2.0 and 2.3.0 on the non-tagged commits. It would have been better if the comment update was made within the commit that changed the functionality. I think I would not put information about non-tagged releases within the code comments as it could be downloaded as a zip or approached by someone without a software background for consumption.

Maybe more appropriate would be wording saying starting with version 2.3.0?

DanRStevens:
Yeah, that might be more clear.

PR #300, opened 2020-06-12
Finalize release3.0.0

### LINKED ISSUE #138: Update documentation before next release
We need to update documentation before new release to include: 

* new names BuiltInModules and ExternalModules
* The fact that ExternalModules have 2 subtypes, ConsoleModules and IniModules
* How BuiltInModules are loaded
* New behaviour of Console Modules that do not replace volumes, but load same named volumes before the game does

Places to look at updating:
* Readme
* SampleModules folder
* op2ext.h (public include)
* TestModule comments

Brett208:
Closes #138 

@DanRStevens, I would be okay with you taking a second pass on the documentation if you want to.

Once this branch is merged, I'll post the official release.


Brett208:
OK, I think this resolves the concerns brought up.

testDll/op2ext.test.cpp
```diff
@@ -5,6 +5,8 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 
+#include "../SampleModules/ConsoleModuleSample.cpp"
+#include "../SampleModules/IniModuleSample.cpp"
```
DanRStevens, line 9: Well that works. Just want to point out this is including .cpp files. If they had defined functions with the same names, it would lead to violations of the one definition rule. In this case, despite both files having corresponding functions, they are each named according to different conventions. As long as things stay simple though, this is probably the fastest and easiest way to get those files compiler checked. I was expecting this to require a more involved change, such as new project files.

Brett208: Yeah, maybe in the next release it can be cleaned up. I'm not even sure those sample files are even useful?

DanRStevens: Yeah, can always be cleaned up later if we want to do it another way.

Not sure about the usefulness. Were these samples being used in testing? Either way, probably makes sense to provide a sample or basic template. I know we have some actual code we can direct people to, but that might be less focused and harder to start from.

PR #302, opened 2020-06-15
Updated release build to only include op2ext.dll and op2ext.pdb

### LINKED ISSUE #301: Automated Release Tagging Pulls too Many Artifacts
The automated release artifact pull included the following files:

* op2ext.dll (Only file that is needed)
* op2mod.dll
* test.exe
* testDll.exe

The debug symbol file include:

* gtest.pdb
* gtest_main.pdb
* op2ext.pdb (only file that I think is needed)
* op2mod.pdb
* test.pdb
* testDll.pdb

Will plan to manually update release zips late today. Need to follow-up with fixing appveyor artifact transfer.

Brett208:
Still allow all artifacts for non-release tagged releases.

There is room to exclude the two gtest pdb files as well from the non-release tag artifacts, but maybe ok if they are present.

Closes #301


Brett208:
Ok. I had thought maybe you were using the test module exe for testing, but I'll pare it down tomorrow. Thanks for looking at it.

Brett208:
Force push to ammend original commit.

